PR #12, opened 2017-06-23
Generate .proto files from Freestyle service protocols

raulraja:
Generate .proto files from Freestyle service protocols to quickly bootstrap your RPC services and have the runtime work standalone with existing proto files. Typeclass based behaviors easily overridable once we turn into into part of the sbt plugin.

Running this:
```bash
sbt
rpc/run "/home/raulraja/Documents/workspace/freestyle-rpc/demo/protocolgen/src/main/scala/service.scala" "/home/raulraja/Documents/workspace/freestyle-rpc/demo/protocolgen/src/main/scala/codegen/"
```
on this service definition: 
_service.scala_
```scala
@option(name = "java_package", value = "com.example.foo", quote = true)
@option(name = "java_multiple_files", value = "true", quote = false)
@option(name = "java_outer_classname", value = "Ponycopter", quote = true)
object protocols {

  @message
  case class MessageRequest(name: String, n: Option[Int])

  @message
  case class MessageReply(name: String, n: List[Int])

  @free
  @service
  trait GreetingService {

    @rpc def sayHello(msg: MessageRequest): FS[MessageReply]

    @rpc
    @stream[ResponseStreaming]
    def lotsOfReplies(msg: MessageRequest, observer: StreamObserver[MessageReply]): FS[Unit]

    @rpc
    @stream[RequestStreaming]
    def lotsOfGreetings(
        @stream msg: StreamObserver[MessageReply]): FS[StreamObserver[MessageRequest]]

    @rpc
    @stream[BidirectionalStreaming]
    def bidiHello(@stream msg: StreamObserver[MessageReply]): FS[StreamObserver[MessageRequest]]
  }

}
```
Generates this `.proto` file:
```proto
syntax = "proto3";

option java_package = "com.example.foo";
option java_multiple_files = true;
option java_outer_classname = "Ponycopter";

message MessageRequest {
   string name = 1;
   int32 n = 2;
}
           
message MessageReply {
   string name = 1;
   int32 n = 2;
}
           
service GreetingService {
   rpc sayHello (MessageRequest) returns (MessageReply) {}
   rpc lotsOfReplies (MessageRequest) returns (stream MessageReply) {}
   rpc lotsOfGreetings (stream MessageReply) returns (MessageRequest) {}
   rpc bidiHello (stream MessageReply) returns (stream MessageRequest) {}
}
```
Before merging this I plan on adding support for `repeated`, `optional` and `required` params mods and tests for each one of the typeclasses.

### DIFF
--- a/demo/protocolgen/src/main/scala/service.scala
+++ b/demo/protocolgen/src/main/scala/service.scala
@@ -7,24 +7,34 @@ import freestyle.internal.EffectLike
 import freestyle.rpc.protocol._
 import io.grpc.stub.StreamObserver
 
-@message
-case class MessageRequest(name: String, n: Int)
+@option(name = "java_package", value = "com.example.foo", quote = true)
+@option(name = "java_multiple_files", value = "true", quote = false)
+@option(name = "java_outer_classname", value = "Ponycopter", quote = true)
+object protocols {
 
-@message
-case class MessageReply(name: String, n: Int)
+  @message
+  case class MessageRequest(name: String, n: Int)
 
-@service
-trait GreetingService[F[_]] extends EffectLike[F] {
+  @message
+  case class MessageReply(name: String, n: Int)
 
-  @rpc def sayHello(msg: MessageRequest): FS[MessageReply]
+  @service
+  trait GreetingService[F[_]] extends EffectLike[F] {
 
-  @rpc @stream[ResponseStreaming]
-  def lotsOfReplies(msg: MessageRequest, observer: StreamObserver[MessageReply]): FS[Unit]
+    @rpc def sayHello(msg: MessageRequest): FS[MessageReply]
 
-  @rpc @stream[RequestStreaming]
-  def lotsOfGreetings(
-      @stream msg: StreamObserver[MessageReply]): FS[StreamObserver[MessageRequest]]
+    @rpc
+    @stream[ResponseStreaming]
+    def lotsOfReplies(msg: MessageRequest, observer: StreamObserver[MessageReply]): FS[Unit]
 
-  @rpc @stream[BidirectionalStreaming]
-  def bidiHello(@stream msg: StreamObserver[MessageReply]): FS[StreamObserver[MessageRequest]]
+    @rpc
+    @stream[RequestStreaming]
+    def lotsOfGreetings(
+        @stream msg: StreamObserver[MessageReply]): FS[StreamObserver[MessageRequest]]
+
+    @rpc
+    @stream[BidirectionalStreaming]
+    def bidiHello(@stream msg: StreamObserver[MessageReply]): FS[StreamObserver[MessageRequest]]
+  }
+
 }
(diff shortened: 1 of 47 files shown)


demo/protocolgen/src/main/scala/service.scala
```diff
@@ -0,0 +1,40 @@
...
+@option(name = "java_outer_classname", value = "Ponycopter", quote = true)
+object protocols {
+
+  @message
+  case class MessageRequest(name: String, n: Int)
+
+  @message
+  case class MessageReply(name: String, n: Int)
+
+  @service
```
juanpedromoreno: is `@service` annotation also generating (automatically) the algebra? Or, should the user annotate the service with `@free` tag?

rpc/src/main/scala/ProtoCodeGen.scala
```diff
@@ -0,0 +1,62 @@
...
+import cats.implicits._
+import freestyle.rpc.protocol.encoders._
+import freestyle.rpc.protocol.model._
+import freestyle.rpc.protocol.processors._
+
+import scala.util.{Failure, Success, Try}
+
+object ProtoCodeGen {
+
+  def main(args: Array[String]): Unit = {
```
juanpedromoreno: I'd suggest removing debug and commented code in this file.

rpc/src/main/scala/proto/converters.scala
```diff
@@ -0,0 +1,194 @@
...
+      }
+  }
+
+  trait ScalaMetaObject2ProtoOptions {
+    def convert(o: Object): List[ProtoOption]
+  }
+
+  object ScalaMetaObject2ProtoOptions {
+    implicit def defaultObject2Options: ScalaMetaObject2ProtoOptions =
+      new ScalaMetaObject2ProtoOptions {
```
juanpedromoreno: It could be simplified to

```scala
object ScalaMetaObject2ProtoOptions {
    implicit def defaultObject2Options: ScalaMetaObject2ProtoOptions =
      (o: Object) => {
        o.mods.collect {
          case Mod.Annot(
              Term.Apply(
                Ctor.Ref.Name("option"),
                Seq(
                  Term.Arg.Named(Term.Name("name"), Lit.String(name)),
                  Term.Arg.Named(Term.Name("value"), Lit.String(value)),
                  Term.Arg.Named(Term.Name("quote"), Lit.Boolean(quote))))) =>
            ProtoOption(name, value, quote)
        }.toList
      }
  }
```

rpc/src/main/scala/proto/converters.scala
```diff
@@ -0,0 +1,194 @@
...
+      }
+  }
+
+  trait ScalaMetaParam2ProtoMessageField {
+    def convert(p: Term.Param, tag: Int): ProtoMessageField
+  }
+
+  object ScalaMetaParam2ProtoMessageField {
+    implicit def defaultParam2ProtoMessageField: ScalaMetaParam2ProtoMessageField =
+      new ScalaMetaParam2ProtoMessageField {
```
juanpedromoreno: It could be just:

```scala
object ScalaMetaParam2ProtoMessageField {
    implicit def defaultParam2ProtoMessageField: ScalaMetaParam2ProtoMessageField =
      (p: Term.Param, tag: Int) => p match {
        case param"..$mods $paramname: Double = $expropt" =>
          ProtoDouble(name = paramname.value, tag = tag)
             ...
```

PR #13, opened 2017-06-26
Provides grpc configuration DSL and GrpcServer algebras

juanpedromoreno:
Server Demo has been also adapted to the newly introduced changes, where the RPC core provides helpers in order to bootstrap the server in the runtime monad, according to the server configuration.

This is a preliminary version towards to fulfill https://github.com/frees-io/freestyle-rpc/issues/10. We still need to find the way to aggregate all the services which would run in such a server.

For now, the user has to provide an implicit list with all the services the server will register:

https://github.com/frees-io/freestyle-rpc/compare/feature/server-definitions#diff-09bc282791d5598d8a70f248fbb125b0R33

Thanks @peterneyens for your suggestions related to the `GrpcConfigInterpreter` :)

Server bootstrap:

```scala
import cats.implicits._
import runtime.implicits._
import freestyle.rpc.server._
import freestyle.rpc.server.implicits._

import scala.concurrent.ExecutionContext.Implicits.global
import scala.concurrent.duration.Duration
import scala.concurrent.Await

object GreetingServerApp {

  def main(args: Array[String]): Unit =
    Await.result(server[GrpcServer.Op].bootstrapFuture, Duration.Inf)
}
```

(`bootstrapFuture` would be equivalent to `bootstrap[Future]`)

Runtime dependencies:

```scala
import freestyle._
import freestyle.implicits._
import freestyle.rpc.demo.greeting._
import freestyle.rpc.server._
import freestyle.rpc.server.implicits._
import freestyle.rpc.server.handlers._

import scala.concurrent.{ExecutionContext, Future}

object implicits {

  implicit val config = Config(portNode1)

  implicit val grpcConfigs: List[GrpcConfig] = List(
    AddService(GreeterGrpc.bindService(new GreetingService, ExecutionContext.global))
  )

  implicit val grpcServerKleisli =
    new GrpcServerHandler[Future] andThen new GrpcConfigInterpreter[Future]

}
```

### DIFF
--- a/rpc/src/main/scala/server/GrpcConfig.scala
+++ b/rpc/src/main/scala/server/GrpcConfig.scala
@@ -24,9 +24,9 @@ import io.grpc._
 
 case class Config(port: Int)
 
-sealed trait GrpcConfig
+sealed trait GrpcConfig extends Product with Serializable
 
-case object GetDirectExecutor extends GrpcConfig
+case object DirectExecutor extends GrpcConfig
 
 case class SetExecutor(executor: Executor) extends GrpcConfig
 
--- a/rpc/src/main/scala/server/GrpcConfig.scala
+++ b/rpc/src/main/scala/server/GrpcConfig.scala
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2017 47 Degrees, LLC. <http://www.47deg.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package freestyle.rpc
+package server
+
+import java.io.File
+import java.util.concurrent.Executor
+
+import io.grpc._
+
+case class Config(port: Int)
+
+sealed trait GrpcConfig
+
+case object GetDirectExecutor extends GrpcConfig
+
+case class SetExecutor(executor: Executor) extends GrpcConfig
+
+case class AddService(service: ServerServiceDefinition) extends GrpcConfig
+
+case class AddBindableService(bindableService: BindableService) extends GrpcConfig
+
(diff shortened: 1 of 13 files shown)


rpc/src/main/scala/server/implicits.scala
```diff
@@ -0,0 +1,58 @@
...
+
+import cats.{~>, Monad}
+import freestyle.{Capture, FreeS}
+import freestyle._
+import freestyle.implicits._
+
+import scala.concurrent.Future
+
+trait FutureCaptureInstance {
+  implicit val `scala.concurrent.FutureCaptureInstance`: Capture[Future] = new Capture[Future] {
```
peterneyens: It looks like there is a `Capture[Future]` instance in the `Capture` companion object.

rpc/src/main/scala/server/handlers/GrpcServerHandler.scala
```diff
@@ -0,0 +1,67 @@
...
+      override def run(): Unit = {
+        shutdown
+        (): Unit
+      }
+    })
+
+    Kleisli(s => C.capture(s.start()))
+  }
+
+  def getPort: GrpcServerOps[F, Int] = Kleisli(s => C.capture(s.getPort))
```
peterneyens: Maybe we can create a small helper method for this `Kleisli(s => C.capture(s.foo))` pattern?

Something like

```scala
private def captureWithServer[A](f: Server => A] : GrpcServerOps[F, A] = Kleisli(s => C.capture(f(s))`

def getPort: GrpcServerOps[F, Int] = captureWithServer(_.getPort)
```

rpc/src/main/scala/server/GrpcConfig.scala
```diff
@@ -0,0 +1,47 @@
...
+import java.io.File
+import java.util.concurrent.Executor
+
+import io.grpc._
+
+case class Config(port: Int)
+
+sealed trait GrpcConfig
+
+case object GetDirectExecutor extends GrpcConfig
```
peterneyens: I haven't look too closely at the `grpc` api, but `GetDirectExecutor` seems a bit strange for something which will never return something?

rpc/src/main/scala/server/GrpcConfig.scala
```diff
@@ -0,0 +1,47 @@
...
+package server
+
+import java.io.File
+import java.util.concurrent.Executor
+
+import io.grpc._
+
+case class Config(port: Int)
+
+sealed trait GrpcConfig
```
peterneyens: We should probably add `extends Product with Serializable`, otherwise the type of something like `List(GetDirectExecutor)` would be inferred as `List[GrpcConfig with Product with Serializable]`.

PR #16, opened 2017-06-28
Client Definitions based on free algebras - Unary Services 

juanpedromoreno:
This PR solves partially the issue https://github.com/frees-io/freestyle-rpc/issues/11  (only unary services).

It brings some free algebras on top of Grpc for RPC clients. Concretely, it solves:

* Channel Configuration.
* Channel Algebra to be able to create new RPC calls, based on method descriptors and some call options.
* ClientCalls Algebra to perform the different types of calls (unary, client streaming, server streaming and bidirecctional).

It also updates the demo code to use these algebras, composing RPC services as freestyle modules:

```scala
@module
trait ClientAPP {
  val greetingClientM: GreetingClientM
  val echoClientM: EchoClientM
}

object GreetingClientApp {

  val messageRequest = MessageRequest("Freestyle")
  val echoRequest    = EchoRequest("echo...")

  def unaryDemo[F[_]](implicit APP: ClientAPP[F]): FreeS[F, (String, String, String)] = {

    val greetingClientM: GreetingClientM[F] = APP.greetingClientM
    val echoClientM: EchoClientM[F]         = APP.echoClientM

    val defaultOptions = CallOptions.DEFAULT

    for {
      hi   <- greetingClientM.sayHello(messageRequest, defaultOptions)
      bye  <- greetingClientM.sayGoodbye(messageRequest, defaultOptions)
      echo <- echoClientM.echo(echoRequest, defaultOptions)
    } yield {
      println(s"Received -> (${hi.message}, ${bye.message}, ${echo.message})")
      (hi.message, bye.message, echo.message)
    }
  }

  def runProgram[F[_]](implicit M: Monad[F]) = {

    Await.result(unaryDemo[ClientAPP.Op].interpret[Future], Duration.Inf)
  }
}
```

### DIFF
--- a/demo/greeting/src/main/scala/greeting/GreetingClientApp.scala
+++ b/demo/greeting/src/main/scala/greeting/GreetingClientApp.scala
@@ -22,7 +22,7 @@ import cats.implicits._
 import freestyle._
 import freestyle.implicits._
 import freestyle.rpc.demo.echo_messages._
-import runtime.implicits.client._
+import runtime.client.implicits._
 import greeting.client._
 import io.grpc._
 
@@ -37,27 +37,32 @@ trait ClientAPP {
 
 object GreetingClientApp {
 
+  type UnaryDemoResponse = (MessageReply, MessageReply, EchoResponse)
+
   val messageRequest = MessageRequest("Freestyle")
   val echoRequest    = EchoRequest("echo...")
 
-  def unaryDemo[F[_]](implicit APP: ClientAPP[F]): FreeS[F, (String, String, String)] = {
+  def unaryDemo[F[_]](implicit APP: ClientAPP[F]): FreeS[F, UnaryDemoResponse] = {
 
     val greetingClientM: GreetingClientM[F] = APP.greetingClientM
     val echoClientM: EchoClientM[F]         = APP.echoClientM
 
     val defaultOptions = CallOptions.DEFAULT
 
-    for {
-      hi   <- greetingClientM.sayHello(messageRequest, defaultOptions)
-      bye  <- greetingClientM.sayGoodbye(messageRequest, defaultOptions)
-      echo <- echoClientM.echo(echoRequest, defaultOptions)
-    } yield {
-      println(s"Received -> (${hi.message}, ${bye.message}, ${echo.message})")
-      (hi.message, bye.message, echo.message)
+    val tupled = (
+      greetingClientM.sayHello(messageRequest, defaultOptions) |@|
+        greetingClientM.sayGoodbye(messageRequest, defaultOptions) |@|
+        echoClientM.echo(echoRequest, defaultOptions)
+    ).tupled
+
+    tupled.map {
+      case (hi: MessageReply, bye: MessageReply, echo: EchoResponse) =>
+        println(s"Received -> (${hi.message}, ${bye.message}, ${echo.message})")
+        (hi, bye, echo)
     }
   }
 
-  def runProgram[F[_]](implicit M: Monad[F]) = {
+  def runProgram[F[_]](implicit M: Monad[F]): Unit = {
 
     Await.result(unaryDemo[ClientAPP.Op].interpret[Future], Duration.Inf)
 
--- a/demo/greeting/src/main/scala/greeting/GreetingClientApp.scala
+++ b/demo/greeting/src/main/scala/greeting/GreetingClientApp.scala
@@ -17,7 +17,6 @@
 package freestyle.rpc.demo
 package greeting
 
-import cats._
 import cats.implicits._
 import freestyle._
 import freestyle.implicits._
@@ -62,7 +61,7 @@ object GreetingClientApp {
     }
   }
 
-  def runProgram[F[_]](implicit M: Monad[F]): Unit = {
+  def main(args: Array[String]): Unit = {
 
     Await.result(unaryDemo[ClientAPP.Op].interpret[Future], Duration.Inf)
 
@@ -88,11 +87,6 @@ object GreetingClientApp {
 
     client.biStreamingDemo()
 
-    (): Unit
-  }
-
-  def main(args: Array[String]): Unit = {
-    runProgram[Future]
     (): Unit
   }
 }
--- a/demo/greeting/src/main/scala/greeting/GreetingClientApp.scala
+++ b/demo/greeting/src/main/scala/greeting/GreetingClientApp.scala
@@ -17,33 +17,57 @@
 package freestyle.rpc.demo
 package greeting
 
-import freestyle.rpc.demo.echo.EchoServiceGrpc
-import freestyle.rpc.demo.echo.EchoServiceGrpc.EchoServiceStub
-import freestyle.rpc.demo.echo_messages.EchoRequest
-import io.grpc.ManagedChannelBuilder
+import cats._
+import cats.implicits._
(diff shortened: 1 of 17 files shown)


demo/greeting/src/main/scala/greeting/GreetingClientApp.scala
```diff
@@ -17,33 +17,57 @@
...
-    // val client  = new GreetingClient(host, portNode2)
+  def unaryDemo[F[_]](implicit APP: ClientAPP[F]): FreeS[F, (String, String, String)] = {
+
+    val greetingClientM: GreetingClientM[F] = APP.greetingClientM
+    val echoClientM: EchoClientM[F]         = APP.echoClientM
+
+    val defaultOptions = CallOptions.DEFAULT
+
+    for {
+      hi   <- greetingClientM.sayHello(messageRequest, defaultOptions)
```
raulraja: Since these operations are not interdependent this could be better expressed in applicative style.
```
(a |@| b |@| c).map( ... )
```

demo/greeting/src/main/scala/greeting/GreetingClientApp.scala
```diff
@@ -17,33 +17,57 @@
...
+      bye  <- greetingClientM.sayGoodbye(messageRequest, defaultOptions)
+      echo <- echoClientM.echo(echoRequest, defaultOptions)
+    } yield {
+      println(s"Received -> (${hi.message}, ${bye.message}, ${echo.message})")
+      (hi.message, bye.message, echo.message)
+    }
+  }
 
-    // http://www.grpc.io/docs/guides/concepts.html
+  def runProgram[F[_]](implicit M: Monad[F]) = {
```
raulraja: Do we need a monad constrain here?

demo/greeting/src/main/scala/greeting/client/GreetingClientM.scala
```diff
@@ -0,0 +1,73 @@
...
+package greeting.client
+
+import freestyle._
+import freestyle.rpc.client._
+import freestyle.rpc.demo.greeting._
+import io.grpc.CallOptions
+import io.grpc.stub.StreamObserver
+
+@module
+trait GreetingClientM {
```
raulraja: :clap:

demo/greeting/src/main/scala/greeting/runtime/implicits.scala
```diff
@@ -40,5 +43,31 @@ object implicits {
...
+
+  implicit def channelMHandler[F[_]]: ChannelM.Op ~> Future =
+    new ChannelMHandler[Future] andThen
+      new ManagedChannelInterpreter[Future](channelFor, channelConfigList)
+
+  implicit val clientCallsMHandler: ClientCallsM.Op ~> Future = new ClientCallsMHandler[Future]
+
+}
+
+object implicits {
```
raulraja: I think users may be confused with this. Every time you import something with the name `implicits` users may go for `implicits._` right away. Since this is a demo is not critical but I think we should try avoiding nesting further implicit scopes once we hit the `implicits` one. For this case I would just have:
`object implicits extends ServerImplicits with ClientImplicits`

rpc/src/main/scala/client/handlers/ClientCallsMHandler.scala
```diff
@@ -0,0 +1,76 @@
...
+import freestyle.Capture
+import freestyle.async.AsyncContext
+import freestyle.rpc.client.ClientCallsM
+import io.grpc.{CallOptions, Channel, ClientCall, MethodDescriptor}
+import io.grpc.stub.{ClientCalls, StreamObserver}
+import freestyle.rpc.client.implicits._
+
+import scala.collection.JavaConverters._
+
+class ClientCallsMHandler[F[_]](implicit C: Capture[F], AC: AsyncContext[F])
```
raulraja: We have traditionally been parametrizing handlers with `M[_]` instead of `F[_]` since in other classes aggregating algebras `F[_]` we have used it to refer to the coproduct. I'm fine if we call this F but in that case we should do this across all modules. Also it would be nice if we all as a team come up with a naming convention doc that we enforce across the entire org to make sure we are consistent and not confusing users.

rpc/src/main/scala/client/handlers/ClientCallsMHandler.scala
```diff
@@ -0,0 +1,76 @@
...
+    extends ClientCallsM.Handler[F] {
+
+  def asyncUnaryCall[I, O](
```
raulraja: Do we actually want to call this things `unary` etc...? Perhaps just `async`, `sync` and `stream` would be enough. thoughts?

rpc/src/main/scala/client/package.scala
```diff
@@ -0,0 +1,66 @@
...
+            case UserAgent(userAgent)               => acc.userAgent(userAgent)
+            case OverrideAuthority(authority)       => acc.overrideAuthority(authority)
+            case UsePlaintext(skipNegotiation)      => acc.usePlaintext(skipNegotiation)
+            case NameResolverFactory(rf)            => acc.nameResolverFactory(rf)
+            case LoadBalancerFactory(lbf)           => acc.loadBalancerFactory(lbf)
+            case SetDecompressorRegistry(registry)  => acc.decompressorRegistry(registry)
+            case SetCompressorRegistry(registry)    => acc.compressorRegistry(registry)
+            case SetIdleTimeout(value, unit)        => acc.idleTimeout(value, unit)
+            case SetMaxInboundMessageSize(max)      => acc.maxInboundMessageSize(max)
+          }).asInstanceOf[ManagedChannelBuilder[_]]
```
raulraja: any way we can get rid of this cast?

rpc/src/main/scala/server/implicits.scala
```diff
@@ -29,8 +29,8 @@ trait Syntax {
 
   final class ServerOps(server: FreeS[GrpcServer.Op, Unit]) {
 
-    def bootstrapM[M[_]](implicit MM: Monad[M], handler: GrpcServer.Op ~> M): M[Unit] =
-      server.interpret[M]
+    def bootstrapM[F[_]](implicit M: Monad[F], handler: GrpcServer.Op ~> F): F[Unit] =
```
raulraja: Regarding the naming conventions mentioned earlier. should this just be `M[_]: Monad` instead of an implicit arg that is not in use in the method body?

demo/greeting/src/main/scala/greeting/GreetingClientApp.scala
juanpedromoreno: Yes, it's required. Not sure if we could skip it somehow

rpc/src/main/scala/client/handlers/ClientCallsMHandler.scala
juanpedromoreno: Sounds good, as we have to differentiate internally (and `@free` doesn't support method overloading), these are the new names: `async`, `asyncStreamServer`, `asyncStreamClient`, `asyncStreamBidi`, `sync`, `syncC`, `syncStreamServer`, `syncStreamServerC`, `asyncM`

rpc/src/main/scala/client/package.scala
juanpedromoreno: Yes

PR #20, opened 2017-07-03
Server/Channel Configuration

### LINKED ISSUE #17: RPC Channel Configuration
We should use `freestyle-config` to parse and load the RPC channel configuration.

juanpedromoreno:
Fixes #17 

This PR adds some logic to be able to load server/client configuration from config files. I'm acquiring a technical debt since for both client/server configurations need to interpret twice. It's something that will be fixed in a separate PR if that makes sense.

### DIFF
--- a/rpc/src/main/scala/client/ChannelConfig.scala
+++ b/rpc/src/main/scala/client/ChannelConfig.scala
@@ -19,6 +19,11 @@ package client
 
 import java.util.concurrent.{Executor, TimeUnit}
 
+import cats.implicits._
+import freestyle._
+import freestyle.config.ConfigM
+import freestyle.config.implicits._
+import freestyle.rpc.client._
 import io.grpc._
 
 sealed trait ManagedChannelFor                               extends Product with Serializable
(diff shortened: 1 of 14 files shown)


rpc/src/main/scala/client/ChannelConfig.scala
```diff
@@ -39,3 +44,24 @@ case class SetDecompressorRegistry(registry: DecompressorRegistry)    extends Ma
...
+
+@module
+trait ChannelConfig {
+
+  val configM: ConfigM
+  val defaultHost = "localhost"
+  val defaultPort = 50051
+
+  def loadChannelAddress(hostPath: String, portPath: String): FS.Seq[ManagedChannelForAddress] =
+    for {
```
raulraja: Can be simplified to `map`

rpc/src/main/scala/client/ChannelConfig.scala
```diff
@@ -39,3 +44,24 @@ case class SetDecompressorRegistry(registry: DecompressorRegistry)    extends Ma
...
+      config <- configM.load
+    } yield
+      ManagedChannelForAddress(
+        config.string(hostPath).getOrElse(defaultHost),
+        config.int(portPath).getOrElse(defaultPort))
+
+  def loadChannelTarget(targetPath: String): FS.Seq[ManagedChannelForTarget] =
+    for {
```
raulraja: Can be simplified to `map`

rpc/src/main/scala/server/ServerConfig.scala
```diff
@@ -15,11 +15,22 @@
...
+@module
+trait ServerConfig {
 
+  val configM: ConfigM
+
+  val defaultPort = 50051
+
+  def loadConfigPort(portPath: String): FS.Seq[Config] =
+    for {
+      config <- configM.load
```
raulraja: Can be simplified to `map`

PR #23, opened 2017-07-05
Adds additional server definitions tests

juanpedromoreno:
This PR completes https://github.com/frees-io/freestyle-rpc/issues/15

Now, rpc server module reaches almost the 100% of test coverage. In the next PRs I'll cover the client definition tests.

Please, @fedefernandez could you take a look? Thanks!

### DIFF
--- a/rpc/src/test/scala/server/ServerConfigTests.scala
+++ b/rpc/src/test/scala/server/ServerConfigTests.scala
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2017 47 Degrees, LLC. <http://www.47deg.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package freestyle.rpc
+package server
+
+import cats.Id
+import cats.data.Kleisli
+import cats.implicits._
+import freestyle._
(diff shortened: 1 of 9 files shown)


rpc/src/test/scala/server/ServerConfigTests.scala
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+  }
+
+  "SServerBuilder" should {
+
+    "work as expected" in {
+
+      val configList: List[GrpcConfig] = List(AddService(sd1))
+      val server: Server               = SServerBuilder(port).withGrpcConfigList(configList).buildServer
```
fedefernandez: Why not just:
```scala
SServerBuilder(port, configList)
```

rpc/src/test/scala/server/ServerConfigTests.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  implicit val ec: ExecutionContext = ExecutionContext.Implicits.global
+
+  "ServerConfig.Op" should {
+
+    "load the port specified in the config file" in {
+
+      val serverF = BuildServerFromConfig[ServerConfig.Op]("rpc.server.port").interpret[Future]
+
+      serverF map (_.getPort shouldBe port)
```
fedefernandez: Because this is a `Future`, this test is not verifying the value. I've tried different values for `port` and it succeeds for all of them

rpc/src/test/scala/server/ServerConfigTests.scala
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    "load the default port when the config port path is not found" in {
+
+      val serverF = BuildServerFromConfig[ServerConfig.Op]("rpc.wrong.path").interpret[Future]
```
fedefernandez: Same as above

rpc/src/test/scala/server/ServerConfigTests.scala
juanpedromoreno: Because this test is covering `withGrpcConfigList` branch ;)

fedefernandez: :+1:

Issue #24, opened 2017-07-06
Test Coverage for Client Definitions


juanpedromoreno:
The purpose of this ticket is to cover the client definitions with new unit tests.

PR #27, opened 2017-07-06
Adds tests for some client handlers

juanpedromoreno:
Also, restructures the test suite classes to use them across client and server tests.

Related to #24

### DIFF
--- a/rpc/src/test/scala/RpcBaseTestSuite.scala
+++ b/rpc/src/test/scala/RpcBaseTestSuite.scala
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2017 47 Degrees, LLC. <http://www.47deg.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


rpc/src/test/scala/RpcBaseTestSuite.scala
```diff
@@ -0,0 +1,49 @@
...
+import cats.data.Kleisli
+import org.scalamock.scalatest.MockFactory
+import org.scalatest.{Matchers, OneInstancePerTest, WordSpec}
+
+import scala.concurrent.duration.Duration
+import scala.concurrent.{Await, Future}
+
+trait RpcBaseTestSuite extends WordSpec with Matchers with OneInstancePerTest with MockFactory {
+
+  sealed trait Syntax {
```
fedefernandez: Why `sealed`?

juanpedromoreno: It's not a big thing, I marked as sealed because it's already associated with the Helper trait below, being sealed it's not composable in other ways outside of its container.

PR #45, opened 2017-09-08
Makes the ChannelBuilder build a public method

juanpedromoreno:
This PR removes the unnecessary `private` modifier for the `Channel` build method.

@vejeta ^^


vejeta:
LGTM

PR #50, opened 2017-10-05
Adds warning about generated proto files

juanpedromoreno:
Solves https://github.com/frees-io/freestyle-rpc/issues/47

It also fixes:

* Skip all of those scala files which don't contain any service/protocol definitions
* Moves back to rpc core the `ProtoCodeGen`, and its deps.

Finally, it suggests a new release, `0.0.4`.


rpc/src/main/scala/proto/model.scala
```diff
@@ -193,7 +193,10 @@ object model {
   final case class ProtoOption(name: String, value: String, quote: Boolean)
 
   final case class ProtoDefinitions(
-      prelude: String = "syntax = \"proto3\";",
+      prelude: String = """|// This file has been automatically generated for use by
```
raulraja: Add link to frees.io future doc section

juanpedromoreno: 👍 

https://github.com/frees-io/freestyle-rpc/issues/51

PR #52, opened 2017-10-06
Fixes Travis Builds

juanpedromoreno:
This PR causes Travis to run the `script` and` afterSuccess` sections separately, one for the rpc kernel, one for the add-in.


.travis.yml
```diff
@@ -1,14 +1,12 @@
 language: scala
 
-sudo: required
+jdk:
+- oraclejdk8
 
 scala:
 - 2.11.11
 - 2.12.3
```
fedefernandez: Maybe we could define an [environment variable](https://docs.travis-ci.com/user/environment-variables/#Defining-public-variables-in-.travis.yml) for the `2.12.3` value.

PR #60, opened 2017-10-19
Test Coverage Server Definitions

juanpedromoreno:
This PR increases the test coverage to ~90%, including test cases able to cover almost all the service and client calls, together with MonixAdapters code.


rpc/src/test/scala/Utils.scala
```diff
@@ -0,0 +1,320 @@
...
+  object clientProgram {
+
+    import service._
+
+    @free
+    trait MyRPCClient {
+      def u(x: Int, y: Int): FS[C]
+      def ss(a: Int, b: Int): FS[List[C]]
+      def cs(cList: List[C], bar: Int): FS[D]
+      def bs(eList: List[E]): FS[E]
```
fedefernandez: I'm not a big fan of these compressed name

PR #75, opened 2017-11-06
Provides Empty Message

### LINKED ISSUE #63: @rpc Services should allow empty argument list
At this moment, this is not possible

```scala
  @free
  @service
  trait PersonService {

    @rpc def listUsers: FS[List[Person]] // Not allowed

    @rpc def listUsers(foo: Boolean): FS[List[Person]] // Allowed

  }
```

We should update the `@service` macro to offer it.

juanpedromoreno:
As Protobuf provides [Emtpy](https://github.com/google/protobuf/blob/master/src/google/protobuf/empty.proto), this PR provides an empty type: `freestyle.rpc.protocol.Empty` as generic empty message that you can re-use to avoid defining duplicated empty messages in your APIs. A typical example is to use it as the request or the response type of an API method.

For instance:

```scala
@free
@service
trait Foo {      
  @rpc def Bar(empty: Empty): FS[Empty]      
}
```

Fixes #63

### DIFF
--- a/rpc/src/main/scala/proto/annotations.scala
+++ b/rpc/src/main/scala/proto/annotations.scala
@@ -38,6 +38,9 @@ package object protocol {
   class message extends StaticAnnotation
 
   class option(val name: String, val value: String, val quote: Boolean) extends StaticAnnotation
+
+  @message
+  case class Empty()
 }
 
 // $COVERAGE-ON$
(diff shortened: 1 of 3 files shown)


rpc/src/main/scala/proto/annotations.scala
```diff
@@ -38,6 +38,9 @@ package object protocol {
   class message extends StaticAnnotation
 
   class option(val name: String, val value: String, val quote: Boolean) extends StaticAnnotation
+
+  @message
+  case class Empty()
```
raulraja: can this be an `object Empty` instead?

juanpedromoreno: I'm afraid it cannot be. We need a valid type (serializable) that we could use as request/response type.

PR #80, opened 2017-11-09
Async Implicits provided by frees-rpc Implicits

juanpedromoreno:
In order to simplify the use of `frees-rpc`, this PR makes that server and client implicits provide automatically:

* `import freestyle.implicits._`
* `import freestyle.async.implicits._`

Hence, users only need to import either:

* `import freestyle.rpc.client.implicits._`
* `import freestyle.rpc.server.implicits._`


juanpedromoreno:
The RPC ones would include `import freestyle.implicits._`.. I guess users couldn't use it in the standard way.. ambiguity perhaps? 

What does make more sense?

juanpedromoreno:
Users could use the `import freestyle.implicits._ ` if they are not using the Rpc ones.

PR #81, opened 2017-11-12
Builds the marshallers explicitly for method descriptors

juanpedromoreno:
Instead of requiring the marshallers implicitly, this PR is proposing to create them explicitly, when the method descriptors are being created.


juanpedromoreno:
No changes for the user. 

Actually, I'm discarding this PR because the issue I found when testing wasn't finally related to implicits.

Thanks.

Issue #86, opened 2017-11-20
Needed EmptyResponse valid for avro/proto and response/request


eperinan:
`frees-rpc` needs to provide a `NoContent()` of http4s response.

Initially, we could implement it as `case class EmptyResponse(a: Unit)`.

This issue is related to #79 and the optimal solution would be finding out a single solution for both empty requests and responses.

Issue #79, opened 2017-11-07
Empty for Avro


juanpedromoreno:
Currently, `freestyle.rpc.protocol.Empty` is valid for Protocol Buffers, but it is not for Avro.

Using `freestyle.rpc.protocol.Empty` we can found OOM errors like this one:

```bash
[error] (grpc-default-executor-0) java.lang.OutOfMemoryError: GC overhead limit exceeded
[error] java.lang.OutOfMemoryError: GC overhead limit exceeded
[error] 	at org.apache.avro.generic.GenericData.newRecord(GenericData.java:1145)
[error] 	at org.apache.avro.generic.GenericDatumReader.readRecord(GenericDatumReader.java:220)
[error] 	at org.apache.avro.generic.GenericDatumReader.readWithoutConversion(GenericDatumReader.java:174)
[error] 	at org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:152)
[error] 	at org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:144)
[error] 	at com.sksamuel.avro4s.AvroBinaryInputStream.<init>(AvroInputStream.scala:30)
[error] 	at com.sksamuel.avro4s.AvroInputStream$.binary(AvroInputStream.scala:111)
[error] 	at freestyle.rpc.internal.service.encoders$avro$$anon$2.parse(service.scala:271)
[error] 	at io.grpc.MethodDescriptor.parseRequest(MethodDescriptor.java:283)
[error] 	at io.grpc.internal.ServerCallImpl$ServerStreamListenerImpl.messagesAvailable(ServerCallImpl.java:251)
[error] 	at io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener$1MessagesAvailable.runInContext(ServerImpl.java:592)
[error] 	at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
[error] 	at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:107)
[error] 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[error] 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
[error] 	at java.lang.Thread.run(Thread.java:748)
[error] (grpc-default-executor-2) java.lang.OutOfMemoryError: GC overhead limit exceeded
[...]

PR #87, opened 2017-11-20
Case class Empty is valid for Avro as well

eperinan:
It solves #86 #79 

Finally I added the param `s: String = ""` to the existing case class and I created several test for it

### DIFF
--- a/rpc/src/main/scala/proto/annotations.scala
+++ b/rpc/src/main/scala/proto/annotations.scala
@@ -40,7 +40,7 @@ package object protocol {
   class option(val name: String, val value: String, val quote: Boolean) extends StaticAnnotation
 
   @message
-  case class Empty()
+  case class Empty(s: String = "")
 }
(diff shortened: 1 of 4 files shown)


rpc/src/main/scala/proto/annotations.scala
```diff
@@ -40,7 +40,7 @@ package object protocol {
   class option(val name: String, val value: String, val quote: Boolean) extends StaticAnnotation
 
   @message
-  case class Empty()
+  case class Empty(s: String = "")
```
raulraja: what happens if `@message object Empty { val s: String = "" }` ?

fedefernandez: It'll be the same case as [here](https://github.com/frees-io/freestyle-rpc/pull/75#pullrequestreview-74429578), right? But IMHO it deserves a try

PR #91, opened 2017-11-27
Removes non-needed implicits when binding service

juanpedromoreno:
It releases a new patch version.


raulraja:
Is this still valid? seems like CI failed on compilation

juanpedromoreno:
Let's close it for now.

PR #92, opened 2017-11-28
Replace @free with @tagless, and drop the requirement of an annotation

tbrown1979:
This PR drops the `@free` annotation and replaces it with `@tagless`. This PR also removes the requirement that a `@service` be annotated with `@tagless`. You are free to define it with your own parameterized type if you want.

This resulted in some changes to the generated code because we're no longer converting from `Op ~> M` anymore.

### DIFF
--- a/rpc/src/main/scala/internal/service/service.scala
+++ b/rpc/src/main/scala/internal/service/service.scala
@@ -141,7 +141,7 @@ case class ServiceAlg(defn: Defn) extends RPCService {
   }
 
   lazy val serviceBindings: Defn.Def = {
-    val args: Seq[Term.Tuple] = requests.map(c => c.call(""))
+    val args: Seq[Term.Tuple] = requests.map(c => c.call(None))
     q"""
          def bindService[F[_]](implicit algebra: $algName[F], HTask: _root_.freestyle.FSHandler[F, _root_.monix.eval.Task], ME: _root_.cats.MonadError[F, Throwable], C: _root_.cats.Comonad[F], S: _root_.monix.execution.Scheduler): _root_.io.grpc.ServerServiceDefinition =
            new freestyle.rpc.internal.service.GRPCServiceDefBuilder(${Lit.String(algName.value)}, ..$args).apply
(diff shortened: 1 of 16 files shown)


rpc/src/main/scala/internal/service/service.scala
```diff
@@ -32,21 +32,31 @@ object serviceImpl {
...
+  def service(defn: Any): Stat = {
+    defn match {
+      case Term.Block(Seq(cls: Trait, companion: Object)) =>
+        freeServiceExtras(cls, companion)
+      case Term.Block(Seq(cls: Class, companion: Object)) if ScalametaUtil.isAbstract(cls) =>
+        freeServiceExtras(cls, companion)
+      case cls: Trait =>
+        defaultServiceExtras(cls.name, cls)
+      case _ =>
+        abort(s"$invalid. $abstractOnly")
```
tbrown1979: Maybe a better error here since @free isn't required anymore

rpc/src/main/scala/internal/service/service.scala
```diff
@@ -56,60 +66,51 @@ object serviceImpl {
...
 }
 
-case class ServiceAlg(defn: Defn) {
+trait RPCService {
+  import utils._
+
+  def defn: Defn
 
-  val (algName, template) = defn match {
+  lazy val (algName, template) = defn match {
```
tbrown1979: I don't think these all need to be `lazy`.

rpc/src/main/scala/internal/service/service.scala
```diff
@@ -192,33 +221,77 @@ private[internal] case class RPCRequest(
       """
   }
 
-  val call: Term.Tuple = streamingType match {
-    case Some(RequestStreaming) =>
-      q"""
+  def call(callType: Option[String]): Term.Tuple = {
```
tbrown1979: This is sort of a weird way of determining which method to use

rpc/src/main/scala/internal/service/service.scala
```diff
@@ -126,12 +118,38 @@ case class ServiceAlg(defn: Defn) {
...
+          channel: _root_.io.grpc.Channel,
           options: _root_.io.grpc.CallOptions = _root_.io.grpc.CallOptions.DEFAULT)
           (implicit H: FSHandler[_root_.monix.eval.Task, M], E: _root_.scala.concurrent.ExecutionContext) : $clientName[M] =
              new ${clientName.ctorRef(Ctor.Name(clientName.value))}[M](channel, options)
      """
+}
+
+case class ServiceAlg(defn: Defn) extends RPCService {
+  val typeParam: Type.Param = defn match {
+    case c: Class => c.tparams.head
```
tbrown1979: `.head`

juanpedromoreno:
This is great. I'll look at it tomorrow, thanks @tbrown1979!

PR #95, opened 2017-12-04
Server Endpoints and Effect Monad

juanpedromoreno:
Until now, especially the unary endpoint was being executed based on the `apply` method, hence only `scala.concurrent.Future` was a valid Monad for the Server interpretation.

This PR fixes this issue requesting a `cats.effect.Effect` instance in order to be able to suspend all the side effects into the `F` context. Therefore, the servers that currently are interpreting to `scala.concurrent.Future` could move interpreting to `cats.effect.IO`, and then, they can still use the `unsafeToFuture` method in order to evaluate the effect and produce the result in a `scala.concurrent.Future`.

However, the new servers can now interpret to suspendable Monads, like the mentioned `cats.effect.IO` or `monix.eval.Task`.

There are two additional task implemented in this PR:

* Async Instances grouping (`cats.effect.IO`, `monix.eval.Task` and `scala.concurrent.Future`).
* It releases a new version `0.4.1`.

### DIFF
--- a/rpc/src/main/scala/RPCAsyncImplicits.scala
+++ b/rpc/src/main/scala/RPCAsyncImplicits.scala
@@ -31,7 +31,7 @@ import scala.concurrent.duration._
 trait IOCapture {
 
   implicit val ioCapture: Capture[IO] = new Capture[IO] {
-    override def capture[A](a: => A): IO[A] = IO.pure(a)
+    override def capture[A](a: => A): IO[A] = IO(a)
   }
 
 }
--- a/rpc/src/main/scala/RPCAsyncImplicits.scala
+++ b/rpc/src/main/scala/RPCAsyncImplicits.scala
@@ -17,6 +17,7 @@
 package freestyle
 package rpc
 
+import cats.effect.IO
 import cats.{~>, Comonad}
 import freestyle.rpc.client.handlers._
 import journal.Logger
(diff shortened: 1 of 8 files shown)


rpc/src/main/scala/RPCAsyncImplicits.scala
```diff
@@ -27,11 +28,23 @@ import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.{Await, ExecutionContext, Future}
 import scala.concurrent.duration._
 
-trait RPCAsyncImplicits extends freestyle.async.Implicits {
+trait IOCapture {
+
+  implicit val ioCapture: Capture[IO] = new Capture[IO] {
+    override def capture[A](a: => A): IO[A] = IO.pure(a)
```
raulraja: shouldn't this be ?
```scala
override def capture[A](a: => A): IO[A] = IO.suspend(IO.pure(a))
```

rpc/src/main/scala/RPCAsyncImplicits.scala
```diff
@@ -58,19 +83,32 @@ trait RPCAsyncImplicits extends freestyle.async.Implicits {
...
   }
 
+  implicit def task2IO(implicit S: Scheduler): Task ~> IO = new (Task ~> IO) {
+    override def apply[A](fa: Task[A]): IO[A] = fa.toIO
+  }
 }
+
+trait IOAsyncInstances extends BaseAsync {
+
+  implicit val ioComonad: Comonad[IO] = new Comonad[IO] {
```
raulraja: We should get rid of this once we have Streams as first class construct and not nested under `FS[Ob]`

rpc/src/main/scala/RPCAsyncImplicits.scala
raulraja: or just `IO(a)` which also suspends.

rpc/src/main/scala/RPCAsyncImplicits.scala
juanpedromoreno: Looking forward to see it it too, just a little bit of time to tackle it.

PR #96, opened 2017-12-07
Move freestyle rpc to freestyle free rpc

### LINKED ISSUE #469: SPIKE: Investigate about fs2-grpc library
https://github.com/fiadliel/fs2-grpc

We need to know how to build a client and a server and what configuration options provides.

Dependency: https://mvnrepository.com/artifact/org.lyranthe.fs2-grpc/java-runtime_2.12/

AdrianRaFo:
Fixes https://github.com/frees-io/freestyle/issues/469


raulraja:
The RPC and Cassandra lib does not need to be aligned, just have it's macros adjusted if any so it compiles with the new core.

build.sbt
```diff
@@ -28,7 +28,7 @@ lazy val common = project
   .settings(moduleName := "frees-rpc-common")
   .settings(scalacOptions := Seq("-deprecation", "-encoding", "UTF-8", "-feature", "-unchecked"))
 
-lazy val freesV = "0.4.6"
+lazy val freesV = "0.4.7"
```
juanpedromoreno: I guess the new version of Freestyle core will be 0.5.0, due the breaking changes.

rpc/src/main/scala/free/RPCAsyncImplicits.scala
```diff
@@ -14,12 +14,12 @@
  * limitations under the License.
  */
 
-package freestyle
+package freestyle.free
 package rpc
```
juanpedromoreno: Not needed.

raulraja:
Looks like the build needs to be updated to not depend on the tagless module anymore.

juanpedromoreno:
As we had several conflicts (including macros code) I preferred to do it from a fresh branch cut off from the master branch here: https://github.com/frees-io/freestyle-rpc/pull/99

Thanks @AdrianRaFo !

PR #113, opened 2018-01-10
Splits core into Server and Client submodules

juanpedromoreno:
This PR splits up the current code base in different modules, mainly separating `client` and `server`. The main goal is delimiting the dependencies usage since when bootstrapping a server we 
don't want the client dependencies and vice versa.

Additionally, other sub-modules have been created for the sake of simplicity.

* `frees-rpc-common` containing protocol types, with the minimum set of dependencies.
* `frees-rpc-async` containing just the async implicit instances (NTs between effect/async types).
* `frees-rpc-internal` where the macros are placed.
* `frees-rpc-client-core`: algebra and code related to the RPC clients.
* `frees-rpc-client-netty`: it doesn't add any additional code, just a transport layer provider based on `grpc-netty`.
* `frees-rpc-client-okhttp`: similar to the `Netty` one, it doesn't add any additional code, just a transport layer provider based on `grpc-okhttp`.
* `frees-rpc-server`: algebra and code related to the RPC server.
* `frees-rpc-config`: helpers to be able to load the client and the server configuration.
* `frees-rpc-tests`: it won't be published, it just houses tests for client and server altogether.

**Breaking changes**: algebras and handlers `TaskMHandler`, `ChannelM`, `ChannelMHandler` have been removed because they are not needed anymore.


modules/server/src/main/scala/implicits.scala
```diff
@@ -39,6 +40,12 @@ trait Syntax {
   }
 }
 
+trait IOCapture {
```
fedefernandez: I know this is not related to the PR, but why is this called `IOCapture`? Does it make sense to rename it?

PR #117, opened 2018-01-11
frees-rpc Tagless-final Migration - Release 0.8.0

juanpedromoreno:
This PR migrates `frees-rpc` entirely to [tagless-final](http://okmij.org/ftp/tagless-final/index.html) using Freestyle `@tagless`.

Considering all these changes as API breaking, it also releases a new minor version: **0.8.0**.


modules/internal/src/main/scala/client/calls.scala
```diff
@@ -19,7 +19,7 @@ package internal
 package client
 
 import freestyle.async.AsyncContext
-import freestyle.free._
+import freestyle.free.FSHandler
```
fedefernandez: Can we use `~>` instead?

modules/config/src/main/scala/client/ChannelConfig.scala
```diff
@@ -18,24 +18,27 @@ package freestyle.rpc
...
-import freestyle.free._
-import freestyle.free.config.ConfigM
+import freestyle.tagless._
+import freestyle.tagless.config.ConfigM
 
 @module
 trait ChannelConfig {
 
   val configM: ConfigM
+  implicit val functor: Functor
```
fedefernandez: Minor comment, what about putting this at the beginning? 
```scala
implicit val functor: Functor

val configM: ConfigM
val defaultHost: String = "localhost"
val defaultPort: Int    = freestyle.rpc.server.defaultPort
```

modules/server/src/main/scala/implicits.scala
```diff
@@ -17,38 +17,22 @@
...
-trait SyncCapture {
-
-  implicit def syncCapture[F[_]](implicit F: Sync[F]): Capture[F] =
-    new Capture[F] { def capture[A](a: => A): F[A] = F.delay(a) }
-}
-
 trait Helpers {
 
-  def server[M[_]](implicit S: GrpcServer[M]): FreeS[M, Unit] = {
+  def server[F[_]: Monad](implicit S: GrpcServer[F]): F[Unit] = {
```
fedefernandez: Can `Monad` be replaced by `Functor`?

modules/server/src/test/scala/CommonUtils.scala
```diff
@@ -52,17 +47,17 @@ trait CommonUtils {
...
     for {
-      _    <- S.start()
-      port <- S.getPort
+      _ <- S.start()
+      _ <- S.getPort
     } yield ()
   }
 
-  def serverStop[M[_]](implicit S: GrpcServer[M]): FreeS[M, Unit] = {
+  def serverStop[F[_]: Monad](implicit S: GrpcServer[F]): F[Unit] = {
```
fedefernandez: Same as above

modules/server/src/test/scala/CommonUtils.scala
```diff
@@ -52,17 +47,17 @@ trait CommonUtils {
 
   def createServerConf(grpcConfigs: List[GrpcConfig]): ServerW = ServerW(SC.port, grpcConfigs)
 
-  def serverStart[M[_]](implicit S: GrpcServer[M]): FreeS[M, Unit] = {
+  def serverStart[F[_]: Monad](implicit S: GrpcServer[F]): F[Unit] = {
```
fedefernandez: Same as above

modules/server/src/test/scala/withouttagless/Utils.scala
```diff
@@ -75,29 +95,30 @@ object Utils extends CommonUtils {
       import service._
       import freestyle.rpc.protocol._
 
-      class ServerRPCService[F[_]](implicit C: Capture[F], T2F: Task ~> F) extends RPCService[F] {
+      class ServerRPCService[F[_]](implicit F: Applicative[F], T2F: Task ~> F)
```
fedefernandez: `F[_]: Applicative` ?

modules/server/src/test/scala/withtagless/TaglessUtils.scala
```diff
@@ -214,8 +215,10 @@ object TaglessUtils extends CommonUtils {
     // Server Runtime Configuration //
     //////////////////////////////////
 
-    implicit val taglessRPCHandler: TaglessRPCService.Handler[ConcurrentMonad] =
-      new TaglessRPCServiceServerHandler[ConcurrentMonad]
+    implicit def taglessRPCHandler[F[_]](
```
fedefernandez: `F[_]: Applicative` ?

modules/server/src/main/scala/GrpcServer.scala
```diff
@@ -17,34 +17,34 @@
...
-import freestyle.free._
+import freestyle.tagless.tagless
 import io.grpc._
 
 import scala.concurrent.duration.TimeUnit
 
-@free
-trait GrpcServer {
+@tagless
+trait GrpcServer[F[_]] {
```
diesalbla: Why do you want to include the `F`? You can still use the `FS`.

modules/server/src/test/scala/server/GrpcServerTests.scala
```diff
@@ -47,7 +46,7 @@ class GrpcServerTests extends RpcServerTestSuite {
 
     "behaves as expected" in {
 
-      def program[F[_]](implicit APP: GrpcServer[F]): FreeS[F, Result] = {
+      def program[F[_]: Monad](APP: GrpcServer[F]): F[Result] = {
```
fedefernandez: Same as above

modules/server/src/test/scala/server/RpcServerTestSuite.scala
```diff
@@ -64,31 +63,36 @@ trait RpcServerTestSuite extends RpcBaseTestSuite {
...
-        case GetMutableServicesOp()          => serverMock.getMutableServices.asScala.toList
-        case ShutdownOp()                    => serverMock.shutdown()
-        case ShutdownNowOp()                 => serverMock.shutdownNow()
-        case IsShutdownOp()                  => serverMock.isShutdown
-        case IsTerminatedOp()                => serverMock.isTerminated
-        case AwaitTerminationTimeoutOp(t, u) => serverMock.awaitTermination(t, u)
-        case AwaitTerminationOp()            => serverMock.awaitTermination()
-      }
-    }
+    def grpcServerHandlerTests[F[_]](implicit F: Applicative[F]): GrpcServer[F] = {
```
fedefernandez: `F[_]: Applicative`

modules/server/src/main/scala/GrpcServer.scala
juanpedromoreno: Yes, I know. IntelliJ shows less red code in this way.

modules/server/src/main/scala/implicits.scala
peterneyens: For the moment we need `Monad` for the for comprehension below.

We could require `Apply` and write `S.start() *> S.getPort.void`
We could write something similar in the other places where Fede left this comment.

juanpedromoreno: You also need flatMap

modules/server/src/test/scala/server/RpcServerTestSuite.scala
juanpedromoreno: I preferred this way in order to use `F` as `F.pure`, rather than `Applicative[F].pure`.

PR #128, opened 2018-01-12
Mini refactoring of `@service`

peterneyens:
Small clean up

### DIFF
--- a/modules/internal/src/main/scala/service.scala
+++ b/modules/internal/src/main/scala/service.scala
@@ -118,7 +118,7 @@ trait RPCService {
     q"""
        $wartSuppress
        class $clientName[M[_]](channel: _root_.io.grpc.Channel, options: _root_.io.grpc.CallOptions = _root_.io.grpc.CallOptions.DEFAULT)
-          (implicit AC : _root_.freestyle.async.AsyncContext[M], H: _root_.freestyle.free.FSHandler[_root_.monix.eval.Task, M], E: _root_.scala.concurrent.ExecutionContext)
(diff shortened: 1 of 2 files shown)


modules/internal/src/main/scala/service.scala
```diff
@@ -273,24 +262,24 @@ private[internal] object utils {
...
     case q"@rpc($s) @stream[BidirectionalStreaming.type] def $name[..$tparams]($paranName: Observable[$request]): $typeParam[Observable[$response]]" =>
-      Option(RPCRequest(algName, name, utils.serializationType(s), Some(BidirectionalStreaming), request, response))
+      RPCRequest(algName, name, utils.serializationType(s), Some(BidirectionalStreaming), request, response)
     case q"@rpc($s) def $name[..$tparams]($request): $typeParam[$response]" =>
-      Option(RPCRequest(algName, name, utils.serializationType(s), None, paramTpe(request), response))
-    case e =>
-      None
-  }.flatten
+      RPCRequest(algName, name, utils.serializationType(s), None, paramTpe(request), response)
+  }
```
juanpedromoreno: Doing this, @service combined with @tagless might not work, since tagless macro is adding a `mapK` which is not being considered by this @service macro.

peterneyens: In what way did I change anything here?

Isn't `(xs.collect(f): List[Option[B]]).flatten` the same as `xs.collect(g): List[B]`?

juanpedromoreno: Correct, sorry, I missed it :(

PR #138, opened 2018-01-16
gRPC Services Metrics using Prometheus

juanpedromoreno:
This PR adds some new optional artifacts that you might want to add to your build in order to monitor your `gRPC` services.

Concretely, adding:

```
libraryDependencies += "io.frees" %% "frees-rpc-prometheus-server" % "x.y.z"
```

Your RPC services will be able to intercept every RPC call on the server side and collect metrics that you could render in your dashboard, using [Prometheus](https://prometheus.io/).

Adding services without intercepting calls:

```scala
lazy val grpcConfigs: List[GrpcConfig] = List(
  AddService(RPCService.bindService[ConcurrentMonad])
)
```

Intercepting calls with [Prometheus](https://prometheus.io/):

```scala
import freestyle.rpc.interceptors.implicits._

lazy val cr: CollectorRegistry = new CollectorRegistry()
lazy val monitorInterceptor = MonitoringServerInterceptor(configuration.withCollectorRegistry(cr))

lazy val grpcConfigs: List[GrpcConfig] = List(
  AddService(RPCService.bindService[ConcurrentMonad].interceptWith(monitorInterceptor))
)
```

Next steps:

* Support Client Metrics with Prometheus
* Support http://metrics.dropwizard.io/4.0.0/

Inspired by:
* https://github.com/grpc-ecosystem/java-grpc-prometheus
* https://github.com/grpc-ecosystem/go-grpc-prometheus.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -56,6 +56,29 @@ lazy val config = project
   .settings(moduleName := "frees-rpc-config")
   .settings(configSettings)
 
+lazy val interceptors = project
+  .in(file("modules/interceptors"))
+  .settings(moduleName := "frees-rpc-interceptors")
+  .settings(interceptorsSettings)
+
+lazy val `prometheus-shared` = project
+  .in(file("modules/prometheus/shared"))
+  .dependsOn(interceptors % "compile->compile;test->test")
+  .settings(moduleName := "frees-rpc-prometheus-shared")
+  .settings(prometheusSettings)
+
+lazy val `prometheus-server` = project
+  .in(file("modules/prometheus/server"))
+  .dependsOn(`prometheus-shared` % "compile->compile;test->test")
+  .dependsOn(server % "compile->compile;test->test")
+  .settings(moduleName := "frees-rpc-prometheus-server")
+
+lazy val `prometheus-client` = project
+  .in(file("modules/prometheus/client"))
+  .dependsOn(`prometheus-shared` % "compile->compile;test->test")
+  .dependsOn(client % "compile->compile;test->test")
+  .settings(moduleName := "frees-rpc-prometheus-client")
+
 //////////////////////////
 //// MODULES REGISTRY ////
 //////////////////////////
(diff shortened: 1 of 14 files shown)


build.sbt
```diff
@@ -56,6 +56,29 @@ lazy val config = project
...
+lazy val `prometheus-shared` = project
+  .in(file("modules/prometheus/shared"))
+  .dependsOn(interceptors % "compile->compile;test->test")
+  .settings(moduleName := "frees-rpc-prometheus-shared")
+  .settings(prometheusSettings)
+
+lazy val `prometheus-server` = project
+  .in(file("modules/prometheus/server"))
+  .dependsOn(`prometheus-shared` % "compile->compile;test->test")
+  .dependsOn(server % "compile->compile;test->test")
```
fedefernandez: `prometheus-server` only depends on `server` for testing, right? Or the idea is to include `prometheus-server` dependency instead of `server`?

modules/prometheus/server/src/main/scala/MonitoringServerInterceptor.scala
```diff
@@ -0,0 +1,97 @@
...
+case class MonitoringServerCall[Req, Res](
+    serverCall: ServerCall[Req, Res],
+    clock: Clock,
+    serverMetrics: ServerMetricsForMethod)(implicit CFG: Configuration)
+    extends ForwardingServerCall.SimpleForwardingServerCall[Req, Res](serverCall) {
+
+  private[this] val millisPerSecond = 1000L
+
+  private[this] val startInstant: Instant = clock.instant
+  reportStartMetrics()
```
fedefernandez: Is this call to `reportStartMetrics()` intentional? in that case, I think we should add a blank line between the call and the private val

build.sbt
juanpedromoreno: Not only for tests, you need `server` to make it work with `prometheus-server`, that's why it should be added transitively. `prometheus-server` doesn't make sense without `server`.

PR #150, opened 2018-01-23
Now the service requires an Effect instead of AsyncContext and `Task ~> M`

fedefernandez:
With this PR:
* There are no references to `AsyncContext`
* As we're already asking for `Scheduler`, all references to `ExecutionContext` has been replaced by `Scheduler`
* The methods are typed to `Effect`


modules/internal/src/main/scala/client/calls.scala
```diff
@@ -18,26 +18,24 @@ package freestyle.rpc
...
 import org.reactivestreams._
 
-import scala.concurrent.ExecutionContext
-
 object calls {
 
   import freestyle.rpc.internal.converters._
 
-  def unary[M[_], Req, Res](
+  def unary[M[_]: Effect, Req, Res](
```
juanpedromoreno: Let's rename `M` by `F` here too.

modules/internal/src/main/scala/client/calls.scala
```diff
@@ -18,26 +18,24 @@ package freestyle.rpc
...
 import org.reactivestreams._
 
-import scala.concurrent.ExecutionContext
-
 object calls {
 
   import freestyle.rpc.internal.converters._
 
-  def unary[M[_], Req, Res](
+  def unary[F[_]: Effect, Req, Res](
```
raulraja: What requires this to be an Effect? If it's `toIO` and `to[F]` those are in `LiftIO`

peterneyens:
I was thinking about something like this before as well.

The only downside I could think of is that if you use `Task`, you sort of unnecessarily convert back and forward to `IO`.
We could add something like `LiftTask[F[_]]` (in analogy to `LiftIO`), which could save that conversion for `Task`.
I haven't really bench marked in any way how costly such a conversion would be.

The current approach is simpler though. I think we could lessen the constraint from `Effect` to `LiftIO`. Then we should write `F.liftIO(io)` instead of `io.to[F]`.

modules/internal/src/main/scala/client/calls.scala
fedefernandez: For deriving the `AsyncContext` instance with [`frees-async-cats-effect`](https://github.com/frees-io/freestyle/blob/master/modules/async/async-cats-effect/shared/src/main/scala/AsyncCatsEffect.scala#L22-L29)

peterneyens: Just left the same comment 😃 
`toIO` is defined on `Task` directly and requires a `Scheduler`.
If we use `LiftIO.liftIO` instead of `IO#to` we can just use `LiftIO` indeed.

peterneyens: We could change the code there to `Async` instead of `Effect`.

peterneyens: Do we still need `AsyncContext` after this PR?

fedefernandez:
The `LiftTask` is a good idea, I'll give it a try.

About the `Effect` constraint, I think it's not possible right now, the method `listenableFuture2Async` requires an `AsyncContext` that it's derived from the `Effect` with [`frees-async-cats-effect`](https://github.com/frees-io/freestyle/blob/master/modules/async/async-cats-effect/shared/src/main/scala/AsyncCatsEffect.scala#L22-L29)

peterneyens:
Do we still need [these natural transformations](https://github.com/frees-io/freestyle-rpc/blob/master/modules/async/src/main/scala/RPCAsyncImplicits.scala#L28-L39) with this change?

modules/internal/src/main/scala/client/calls.scala
raulraja: @peterneyens at this time `AsyncContext` is there to support `Future` since those can't implement `Sync` but would love to get rid of it if we can support scala Futures in some other way.

peterneyens: Yeah, I forgot that this was at the client side of RPC.

juanpedromoreno: @peterneyens This PR is already removing AsyncContext. These [instances](https://github.com/frees-io/freestyle-rpc/blob/master/modules/async/src/main/scala/RPCAsyncImplicits.scala#L28-L39) can be also removed.

@raulraja We are not supporting `scala.concurrent.Future` before this PR, at least on the server side.

fedefernandez: But we need the `AsyncContext` for the [`listenableFuture2Async`](https://github.com/frees-io/freestyle/blob/master/modules/async/async-guava/src/main/scala/AsyncGuava.scala#L29)

peterneyens: If we change the `Effect` to `Async` [here](https://github.com/frees-io/freestyle/blob/master/modules/async/async-cats-effect/shared/src/main/scala/AsyncCatsEffect.scala#L23), we could probably lower the `Effect` to `Async` in this PR as well.

fedefernandez:
I've added the `LiftTask` (thanks to @peterneyens) to the internal module. Also, the implicit `Task.to[F]` is exposed through the server and client implicitis.

modules/internal/src/main/scala/TaskImplicits.scala
```diff
@@ -15,25 +15,18 @@
...
 import monix.execution.Scheduler
 
-import scala.concurrent.Future
+trait TaskImplicits {
 
-trait RPCAsyncImplicits {
+  implicit class TaskOps[A](task: Task[A]) {
 
-  implicit val future2Task: Future ~> Task = λ[Future ~> Task](Task.deferFuture(_))
+    def to[F[_]](implicit F: LiftIO[F], s: Scheduler): F[A] = effectLiftTask[F].liftTask(task)
```
peterneyens: I would probably take `F: LiftTask[F]` here instead of `LiftIO` and `Scheduler`.

That would probably mean some changes in the macro as well.

modules/server/src/test/scala/withtagless/TaglessUtils.scala
```diff
@@ -146,10 +142,9 @@ object TaglessUtils extends CommonUtils {
       import freestyle.rpc.withtagless.TaglessUtils.client.MyRPCClient
       import freestyle.rpc.protocol._
 
-      class TaglessRPCServiceClientHandler[F[_]: Monad](
+      class TaglessRPCServiceClientHandler[F[_]: Effect](
           implicit client: TaglessRPCService.Client[F],
-          M: MonadError[F, Throwable],
-          T2F: Task ~> F)
+          M: MonadError[F, Throwable])
```
peterneyens: `Effect` already means that this will be available.

modules/internal/src/main/scala/client/calls.scala
fedefernandez: Addressed in https://github.com/frees-io/freestyle/pull/516

fedefernandez:
I've reduced the type constraint to `Async` using https://github.com/frees-io/freestyle/pull/516

This is ready for a new review. Thanks!

juanpedromoreno:
@peterneyens @raulraja is this now ready to go? Thanks.

fedefernandez:
@raulraja I only found the `toIO` method, that I used in the first [commit](https://github.com/frees-io/freestyle-rpc/pull/150/commits/9971302e2ce9bc0ad341dd3fe9feb4574c504106#diff-80abbcbfa27baa15fb86a1f326492576R76) but it would need a double conversion to go to `F`. We've simplified the conversion with the @peterneyens [suggestion](https://github.com/frees-io/freestyle-rpc/pull/150#issuecomment-359767879)

PR #152, opened 2018-01-24
fs2.Stream Support

juanpedromoreno:
This PR brings support for `fs2.Stream[F, ?]` for all the types of streaming:

* Client Streaming
* Server Streaming
* Bi-directional Streaming

This has been possible thanks to @zainab-ali and its great library https://github.com/zainab-ali/fs2-reactive-streams.

From now on, @service's can be defined, for example, as follows:

```scala
import fs2.Stream

// models:
case class A(x: Int, y: Int)
case class B(a1: A, a2: A)
case class C(foo: String, a: A)
case class D(bar: Int)
case class E(a: A, foo: String)

@service
trait RPCService[F[_]] {

  @rpc(Avro) def unary(a: A): F[C]

  @rpc(Protobuf)
  @stream[ResponseStreaming.type]
  def serverStreaming(b: B): Stream[F, C]

  @rpc(Protobuf)
  @stream[RequestStreaming.type]
  def clientStreaming(oa: Stream[F, A]): F[D]

  @rpc(Avro)
  @stream[BidirectionalStreaming.type]
  def biStreaming(oe: Stream[F, E]): Stream[F, E]

}
```

`monix.reactive.Observable` is still supported. The same protocol can be defined, in terms of Observables as follows:

```scala
import monix.reactive.Observable

@service
trait RPCService[F[_]] {

  @rpc(Avro) def unary(a: A): F[C]

  @rpc(Protobuf)
  @stream[ResponseStreaming.type]
  def serverStreaming(b: B): Observable[C]

  @rpc(Protobuf)
  @stream[RequestStreaming.type]
  def clientStreaming(oa: Observable[A]): F[D]

  @rpc(Avro)
  @stream[BidirectionalStreaming.type]
  def biStreaming(oe: Observable[E]): Observable[E]

}
```

**### Breaking Changes**

* @tagless is no longer supported. RPC protocols can be defined entirely with just `@service` annotation.
* For the same reason, protocols don't need to be defined in terms of `F[Observable[?]]` anymore. The same applies to `fs2.Stream`.


juanpedromoreno:
Tests are failing because of random failures that are being fixed at https://github.com/frees-io/freestyle-rpc/pull/153.

Once that it's merged, I'll rebase against master.

PR #153, opened 2018-01-24
Try to fix the random test failure in client metrics

fedefernandez:
Fixes the random test failure by retrying the *completed* metrics check until a max retry is reached or the check pass.


fedefernandez:
Merging this to solve the test failure related with f2s in the feature branch

PR #159, opened 2018-02-02
Refactor service macro

peterneyens:
Played around a bit to get a better understanding what is actually happening in `@service`.
I think this cleans the code up a bit.


peterneyens:
I didn't disable coverage for the marshallers after moving them to a separate file.

juanpedromoreno:
No problem, it's ok, those are pending to be implemented.

PR #161, opened 2018-02-06
Allows adding compression at method level

fedefernandez:
As per [this comment](https://github.com/grpc/grpc-java/issues/2196#issuecomment-242141285) seems that this is the right way for enabling compression on the server.

@juanpedromoreno thoughts?

### DIFF
--- a/modules/common/src/main/scala/protocol.scala
+++ b/modules/common/src/main/scala/protocol.scala
@@ -32,7 +32,12 @@ sealed trait SerializationType extends Product with Serializable
 case object Protobuf           extends SerializationType
 case object Avro               extends SerializationType
 
-class rpc(val serializationType: SerializationType)                   extends StaticAnnotation
+sealed abstract class CompressionType extends Product with Serializable
+case object Identity                  extends CompressionType
+case object Gzip                      extends CompressionType
+
+class rpc(val serializationType: SerializationType, val compressionType: CompressionType = Identity)
+    extends StaticAnnotation
 class stream[S <: StreamingType]                                      extends StaticAnnotation
 class message                                                         extends StaticAnnotation
 class option(val name: String, val value: String, val quote: Boolean) extends StaticAnnotation
--- a/modules/internal/src/main/scala/server/monixCalls.scala
+++ b/modules/internal/src/main/scala/server/monixCalls.scala
@@ -34,41 +34,54 @@ object monixCalls {
 
   import freestyle.rpc.internal.converters._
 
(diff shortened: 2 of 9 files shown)


modules/common/src/main/scala/protocol.scala
```diff
@@ -32,7 +32,12 @@ sealed trait SerializationType extends Product with Serializable
 case object Protobuf           extends SerializationType
 case object Avro               extends SerializationType
 
-class rpc(val serializationType: SerializationType)                   extends StaticAnnotation
+sealed abstract class CompressionType extends Product with Serializable
+case object Identity                  extends CompressionType
```
juanpedromoreno: I'm wondering if we could rename this by something like as `Plain` or similar, what do you think?

juanpedromoreno:
Minor comment added, however, this looks great, thanks @fedefernandez !

modules/common/src/main/scala/protocol.scala
fedefernandez: Yeah, you know, I was trying to be cohesive with [grpc-java](https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/Codec.java) but I agree it's not the best name.

juanpedromoreno: I missed that, forget my suggestion ;)

modules/internal/src/main/scala/server/monixCalls.scala
```diff
@@ -34,41 +34,54 @@ object monixCalls {
...
-  def clientStreamingMethod[F[_]: Effect, Req, Res](f: Observable[Req] => F[Res])(
-      implicit S: Scheduler): ClientStreamingMethod[Req, Res] =
+  def clientStreamingMethod[F[_]: Effect, Req, Res](
+      f: Observable[Req] => F[Res],
+      maybeCompression: Option[String])(implicit S: Scheduler): ClientStreamingMethod[Req, Res] =
     new ClientStreamingMethod[Req, Res] {
 
-      override def invoke(responseObserver: StreamObserver[Res]): StreamObserver[Req] =
+      override def invoke(responseObserver: StreamObserver[Res]): StreamObserver[Req] = {
+        addCompression(responseObserver, maybeCompression)
```
fedefernandez: I'm not very cool with this approach, I was trying to change the `addConmpression` method result to `F[Unit]` but I'm not sure how to add an effect here, in the `transformStreamObserver` that needs to be applied before anything else. I think I'll give it a second try.

PR #162, opened 2018-02-07
SSL/TLS Encryption Support (Netty)

juanpedromoreno:
Among other things, it provides **SSL/TLS** encryption support.

* Server-side example:

```scala

    val serverSslContext: SslContext =
      GrpcSslContexts
        .configure(
          GrpcSslContexts.forServer(serverCertFile, serverPrivateKeyFile),
          SslProvider.OPENSSL)
        .trustManager(serverTrustedCaCerts: _*)
        .clientAuth(ClientAuth.REQUIRE)
        .build()

    val grpcConfigs: List[GrpcConfig] = List(
      SetSslContext(serverSslContext),
      AddService(RPCService.bindService[ConcurrentMonad])
    )

    implicit val serverW: ServerW = ServerW.netty(SC.port, grpcConfigs)

    // Start server...
```

* Client-side example:

```scala
    val clientSslContext: SslContext =
      GrpcSslContexts.forClient
        .keyManager(clientCertChainFile, clientPrivateKeyFile)
        .trustManager(clientTrustedCaCerts: _*)
        .build()

      val channelInterpreter: NettyChannelInterpreter = new NettyChannelInterpreter(
        createChannelFor,
        List(
          OverrideAuthority(TestUtils.TEST_SERVER_HOST),
          NettyUsePlaintext(false),
          NettyNegotiationType(NegotiationType.TLS),
          NettySslContext(clientSslContext)
        )
      )

      val client: RPCService.Client[ConcurrentMonad] =
        RPCService.clientFromChannel[ConcurrentMonad](channelInterpreter.build)

      // Run client...
      client.endpoint(request).unsafeRunSync()
```


project/ProjectPlugin.scala
```diff
@@ -22,6 +22,7 @@ object ProjectPlugin extends AutoPlugin {
       val fs2ReactiveStreams: String = "0.5.0"
       val grpc: String               = "1.9.0"
       val monix: String              = "3.0.0-M3"
+      val netty: String              = "2.0.7.Final"
```
peterneyens: Should we use something different than "netty", it is a tiny bit confusing that is not the netty version, but the version of netty-tcnative-boringssl-static.

modules/server/src/main/scala/GrpcConfig.scala
```diff
@@ -22,7 +22,7 @@ import java.util.concurrent.Executor
 
 import io.grpc._
 
-sealed trait GrpcConfig                                                        extends Product with Serializable
+trait GrpcConfig                                                               extends Product with Serializable
```
peterneyens: Just a question, but what is the reason for unsealing `GrpcConfig`?

juanpedromoreno: The `Netty` settings outside this file should extend `GrpcConfig`, to be able to build it with the rest of the "standard" configurations.

PR #184, opened 2018-02-20
Merge sbt-freestyle-protogen into freestyle-rpc codebase, and update @rpc processing to handle latest `freestyle-rpc syntax

### LINKED ISSUE #180: Sbt Protogen Plugin Migration to the core
https://github.com/frees-io/sbt-freestyle-protogen provides the ability to generate `.proto` files from `frees-rpc` service definitions. It's a useful utility to achieve interoperability between `frees-rpc` and other ecosystems out of Scala through Protocol Buffers. Sadly, now it's completely outdated.

Following the great idea provided by @L-Lavigne, one solution to prevent this issue the next time would be moving back to the core, in order to keep it up to date and additionally, reusing all Scala Meta/Macros code existing in the core.

We’d have a new folder structure for this matter in `frees-rpc`: 

* `/modules/protogen/core/` ~> `frees-idlgen-core` as artifact name perhaps? `frees-idl-core`?
* `/modules/protogen/plugin/` ~> `sbt-frees-idlgen` as artifact name perhaps? `sbt-frees-idl`?

L-Lavigne:
Resolves #180 and fixes https://github.com/frees-io/sbt-freestyle-protogen/issues/24. Note that this initial merge is mostly a straight-up copy of the current `sbt-freestyle-protogen` with different modules and package names, that removes the duplicated macro-processing code for `@rpc`-annotated methods and adds a basic unit test. There's still work to do on refactorings to be prepare for Avro gen, adding test coverage, and fixing some non-standard-compliant `.proto` output. (For the latter case I updated the readme to reflect what `protoGen` actually generates currently). Those remaining issues will be fixed in upcoming patches.

### DIFF
--- a/docs/src/main/tut/README.md
+++ b/docs/src/main/tut/README.md
@@ -416,6 +416,8 @@ addSbtPlugin("io.frees" % "frees-rpc-sbt-idlgen" % "0.11.1")
 ```
 [comment]: # (End Replace)
 
+Note that the plugin is only available for Scala 2.12.
+
 ### Plugin Settings
 
 There are a couple key settings that can be configured according to various needs:
--- a/docs/src/main/tut/README.md
+++ b/docs/src/main/tut/README.md
@@ -412,7 +412,7 @@ Add the following line to _project/plugins.sbt_:
 
 [comment]: # (Start Replace)
 ```scala
-addSbtPlugin("io.frees" % "frees-rpc-sbt-idlgen" % "0.11.1")
+addSbtPlugin("io.frees" % "sbt-frees-rpc-idlgen" % "0.11.1")
 ```
 [comment]: # (End Replace)
 
--- a/docs/src/main/tut/README.md
+++ b/docs/src/main/tut/README.md
@@ -113,6 +113,9 @@ libraryDependencies += "io.frees" %% "frees-rpc-dropwizard-client" % "0.11.1"
 
 // optional - for the communication between server and client by using SSL/TLS.
 libraryDependencies += "io.frees" %% "frees-rpc-netty-ssl" % "0.11.1"
+
+// optional - for .proto file generation.
+libraryDependencies += "io.frees" %% "frees-rpc-sbt-idlgen" % "0.11.1"
 ```
 
 [comment]: # (End Replace)
@@ -152,7 +155,7 @@ Likewise, you can define [gRPC] services in your proto files, with RPC method pa
 // The greeter service definition.
 service Greeter {
   // Sends a greeting
-  rpc SayHello (HelloRequest) returns (HelloReply) {}
+  rpc sayHello (HelloRequest) returns (HelloReply) {}
 }
 
 // The request message containing the user's name.
@@ -201,7 +204,7 @@ import freestyle.rpc.protocol._
 case class Person(name: String, id: Int, has_ponycopter: Boolean)
 ```
 
-As we can see, it’s quite simple since it’s just a Scala case class preceded by the `@message` annotation (`@message` is optional though and used exclusively by [sbt-freestyle-protogen](https://github.com/frees-io/sbt-freestyle-protogen)):
+As we can see, it’s quite simple since it’s just a Scala case class preceded by the `@message` annotation (`@message` is optional though and used exclusively by `protoGen`):
 
 By the same token, let’s see now how the `Greeter` service would be translated to the [frees-rpc] style (in your `.scala` file):
 
@@ -281,13 +284,13 @@ object protocol {
      * @param request Say Hello Request.
      * @return HelloReply.
      */
-    @rpc(Protobuf) def sayHello(request: HelloRequest): F[HelloReply]
+    @rpc(Protobuf) def SayHello(request: HelloRequest): F[HelloReply]
 
-    @rpc(Protobuf) def emptyResponse(request: HelloRequest): F[Empty.type]
+    @rpc(Protobuf) def EmptyResponse(request: HelloRequest): F[Empty.type]
 
-    @rpc(Protobuf) def emptyRequest(request: Empty.type): F[HelloReply]
+    @rpc(Protobuf) def EmptyRequest(request: Empty.type): F[HelloReply]
 
-    @rpc(Protobuf) def emptyRequestRespose(request: Empty.type): F[Empty.type]
+    @rpc(Protobuf) def EmptyRequestRespose(request: Empty.type): F[Empty.type]
   }
 }
 ```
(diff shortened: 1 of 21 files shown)


modules/idlgen/core/src/main/scala/converters.scala
```diff
@@ -0,0 +1,162 @@
...
+    def convert(t: Trait): ProtoService
+  }
+
+  object ScalaMetaTrait2ProtoService {
+    implicit def defaultTrait2ServiceConverter: ScalaMetaTrait2ProtoService =
+      new ScalaMetaTrait2ProtoService {
+        override def convert(t: Trait): ProtoService =
+          ProtoService(
+            t.name.value,
+            ServiceAlg(t).requests.map(
```
L-Lavigne: This is the new hook into the existing macro processing code

docs/src/main/tut/README.md
```diff
@@ -113,6 +113,9 @@ libraryDependencies += "io.frees" %% "frees-rpc-dropwizard-client" % "0.11.1"
 
 // optional - for the communication between server and client by using SSL/TLS.
 libraryDependencies += "io.frees" %% "frees-rpc-netty-ssl" % "0.11.1"
+
+// optional - for .proto file generation.
+libraryDependencies += "io.frees" %% "frees-rpc-sbt-idlgen" % "0.11.1"
```
juanpedromoreno: For the plugin, I think this would be enough: https://github.com/frees-io/freestyle-rpc/pull/184/files#diff-e9a58680801f40c00e3ed7f80bcd7404R418

docs/src/main/tut/README.md
```diff
@@ -152,7 +155,7 @@ Likewise, you can define [gRPC] services in your proto files, with RPC method pa
 // The greeter service definition.
 service Greeter {
   // Sends a greeting
-  rpc SayHello (HelloRequest) returns (HelloReply) {}
+  rpc sayHello (HelloRequest) returns (HelloReply) {}
```
L-Lavigne: This is what the current `protoGen` actually outputs - a fix will address https://github.com/frees-io/sbt-freestyle-protogen/issues/21

modules/idlgen/core/src/main/scala/converters.scala
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  object ScalaMetaParam2ProtoMessageField {
+    implicit def defaultParam2ProtoMessageField: ScalaMetaParam2ProtoMessageField =
+      new ScalaMetaParam2ProtoMessageField {
+        override def convert(
+            p: Term.Param,
+            tag: Int,
+            mod: Option[ProtoFieldMod] = None): ProtoMessageField = p match {
+          case param"..$mods $paramname: Double = $expropt" =>
```
diesalbla: Given that in all of the pattern-matches below we are not using neither the `$paramname` nor the `mods` for choosing cases, we could just match on the `p.decltpe` instead.  
```Scala
p.decltpe match {
    case t"Float" => --- 
    case t"Long" => --- 
}
```
You can get the name of the parameter using `p.name`.

modules/idlgen/core/src/test/scala/ProtoCodeGenTests.scala
```diff
@@ -0,0 +1,43 @@
...
+
+package freestyle.rpc.idlgen.protobuf
+
+import freestyle.rpc.common.RpcBaseTestSuite
+import java.io.File
+import scala.io.Source
+
+class ProtoCodeGenTests extends RpcBaseTestSuite {
+
+  s"$ProtoCodeGen.generate()" should {
```
L-Lavigne: This begins to address https://github.com/frees-io/sbt-freestyle-protogen/issues/18, more to come.
Currently breaks in Travis due to working directory assumption, will fix

docs/src/main/tut/README.md
juanpedromoreno: 👍 Fixing frees-io/sbt-freestyle-protogen#21 in the incoming PRs would be great.

docs/src/main/tut/README.md
L-Lavigne: Ah yes you're right - fixed.

modules/idlgen/core/src/test/scala/ProtoCodeGenTests.scala
L-Lavigne: Made a temporary workaround for testing in Travis, but using the filesystem will not be sustainable so I'll convert the non-plugin parts to use input streams.

L-Lavigne:
Build currently breaks in cross-version testing because `scripted-plugin` doesn't support Scala 2.11 for our version of sbt. Will investigate.

juanpedromoreno:
We could get rid of the Scala 2.11 cross versioning for the SBT plugin, just keeping Scala 2.12.x and Sbt 1.x.y.

modules/idlgen/core/src/main/scala/converters.scala
L-Lavigne: Agreed, I'm working on a refactoring that addresses this and other issues. I think I'll make another PR for it.

modules/idlgen/core/src/main/scala/encoders.scala
```diff
@@ -0,0 +1,113 @@
...
+          s"""service ${a.name} {
+             |${a.rpcs.map(MFEncoder.encode).mkString("  ", "\n  ", "")}
+             |}
+           """.stripMargin
+      }
+
+    implicit def defaultProtoServiceFieldEncoder: ProtoEncoder[ProtoServiceField] =
+      new ProtoEncoder[ProtoServiceField] {
+        override def encode(a: ProtoServiceField): String = a.streamingType match {
+          case None =>
```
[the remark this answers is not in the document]
L-Lavigne: Agreed, I'm working on a refactoring that addresses this and other issues. I think I'll make another PR for it.

modules/idlgen/core/src/main/scala/model.scala
```diff
@@ -0,0 +1,205 @@
...
+    def name: String
+
+    def id: String
+
+    def tag: Int
+
+    def comment: Option[String]
+  }
+
+  final case class ProtoDouble(
```
diesalbla: In this `ProtoMessageField` abstract-data-type, each and every one of the classes below overrides the `def` in the trait with a default parameter `id`. However, ¿is there a reason why there should be an instance of, say, `ProtoFloat` with a value of `id` different to `"float"`?

If that value would always be the same, it could make more sense to use something else than a constructor parameter. For instance, it could be an `override val` inside the body of the class. Or, you could make `ProtoMessageField` into an abstract class with one constructor parameter `id` (or with all fields as constructor parameters).

L-Lavigne: Thanks for reviewing! Agreed that the current code isn't optimal - I left it as-is when migrating from the plugin project so we could focus on the migration/macro-deduplication parts. I have a code refactoring PR ready to push once this is merged, that will address this by using a model that's less tied to Protobufs and will better support Avro once implemented.

PR #207, opened 2018-03-20
Solves encoding issues in docs

eperinan:
This PR solves encoding issues related to `###` in docs.


L-Lavigne:
The diffs look identical - are we changing tabs for spaces or something?

eperinan:
Yes. It's identical, but if you check it in `master` the characters are equal and it fails.

I don't know what happened exactly. I solved it typing these lines again directly from Github.

L-Lavigne:
Interesting. Good find! 👍

PR #210, opened 2018-03-29
Scala source generation from Avro IDL

### LINKED ISSUE #206: Support for generating Scala services from Avro IDL files
Following the Protobuf IDL generation migration in #184 and Avro IDL generation support in #195, we'll want to support going in the other direction and generating Scala services from the IDL. This could mitigate some binary compatibility issues when upgrading services. We'll start with Avro.

L-Lavigne:
Closes #206.

Summary of changes:
- New plugin task `srcGen` that generates Freestyle-RPC-annotated Scala sources from IDL defintions (currently only Avro).
- Avro implementation is mostly a wrapper around AvroHugger, with special treatment of RPC requests.
- Supports generating from both `.avdl` (IDL) and `.avpr` (JSON). Note that currently the Avro IDL generator only generates `.avpr` which is Avro's "real" internal format, but `.avdl` has the advantage of allowing imports.
- Can be used either as a one-time task or a as a compiler plugin.
- Both `idlGen` and `srcGen` now require the IDL type (either `proto` or `avro`) as their first parameter instead of always trying to generate both.
- Both tasks also accept "options" as tail arguments. Currently those are used by `srcGen avro` as `@rpc` arguments, which currently include only `Gzip`.
- The task options `sourceDir` and `targetDir` are renamed `idlGenSourceDir` and `idlGenTargetDir` for `idlGen`, and `srcGenSourceDir` and `srcGenTargetDir` for `srcGen`. The latter will write to `target/scala-2.12/src_managed` by default, and similarly the default for `idlGenTargetDir` becomes `target/scala-2.12/resource_managed`.
- Internally, most `idlGen` classes were refactored to leverage IDL and Source generator commonalities, and some test utilities were moved to the base suite.
- Documentation was updated for the new plugin behavior.


modules/idlgen/plugin/src/main/scala/IdlGenPlugin.scala
```diff
@@ -28,33 +28,70 @@ object IdlGenPlugin extends AutoPlugin {
     lazy val idlGen: TaskKey[Unit] =
       taskKey[Unit]("Generates IDL files from freestyle-rpc service definitions")
 
-    lazy val sourceDir: SettingKey[File] =
+    lazy val srcGen: TaskKey[Unit] =
```
juanpedromoreno: We might want to make this task as `TaskKey[Seq[File]]`, in order to be able to add the following configuration:

`sourceGenerators in Compile += (srcGen in Compile).taskValue`

triggering the `srcGen` task into the compile phase.

modules/idlgen/plugin/src/main/scala/IdlGenPlugin.scala
```diff
@@ -28,33 +28,70 @@ object IdlGenPlugin extends AutoPlugin {
...
 
   lazy val defaultSettings: Seq[Def.Setting[_]] = Seq(
-    sourceDir := baseDirectory.value / "src" / "main" / "scala",
-    targetDir := baseDirectory.value / "src" / "main" / "resources"
+    idlType := "(missing arg)",
+    idlGenSourceDir := baseDirectory.value / "src" / "main" / "scala",
+    idlGenTargetDir := baseDirectory.value / "src" / "main" / "resources",
+    srcGenSourceDir := baseDirectory.value / "src" / "main" / "resources",
+    srcGenTargetDir := baseDirectory.value / "target" / "scala-2.12" / "src_managed" / "main",
+    genOptions := Seq.empty
```
juanpedromoreno: Could default values be as follows?

```scala
idlGenSourceDir := sourceDirectory.value / "main" / "scala",
idlGenTargetDir := resourceDirectory.value,
srcGenSourceDir := resourceDirectory.value,
srcGenTargetDir := sourceManaged.value,
```

docs/src/main/tut/idl-generation.md
```diff
@@ -161,6 +152,59 @@ And the resulting Avro IDL would be generated in `/src/main/resources/avro/servi
...
+
+  @rpc(Avro)
+  def sayHelloAvro(arg: foo.bar.HelloRequest): F[foo.bar.HelloResponse]
+
+}
+```
+
+You can also integrate this source generation in your compile process by adding this setting to your module:
+```
+sourceGenerators in Compile += (srcGen in Compile).taskValue)
```
juanpedromoreno: 👏

raulraja:
Outstanding work @L-Lavigne :clap: :clap: :clap:

L-Lavigne:
Thanks Raúl!

PR #215, opened 2018-04-04
Avro Messages Serialized With Schemas

tyler-clark:
This PR adds support for serializing avro messages in a format that includes the writer schema in the serialized message. This allows the RPC request and response models to evolve based on [Avro Schema Evolution](https://docs.confluent.io/current/avro.html#schema-evolution) rules without forcing tight dependency coupling between client and server. The existing `Avro` `SerializationType` allows for some very limited model evolutions, but falls short in more complex cases. For example, given an rpc service like
```scala
case class KnownError(value: String)
case class UnknownError(value: String)
case class Success(value: String)
case class Request(value: String)
case class Response(value: KnownError :+: UnknownError  :+: Success :+: CNil)
@service
trait RPCService[F[_]] {
  @rpc(Avro) def get(a: Request): F[Response]
}
```
If my server were to add a field to the `Success` case class, deserialization would fail on any clients until they were able to update their `RPCService.Client` to a version that understands the added field. However, using the `AvroWithSchema` `SerializationType`, no immediate client upgrade would be necessary. Clients would be able to continue to deserialize the message into the model that they understand, and would be free to upgrade at their leisure.

This solution is known to not be the highest performing solution. I think ideally a solution that allows for schema evolutions involves a schema registry, but the changes here I think offer a workable solution in the absence of a schema registry.
I was able to run some benchmarks using [Metrifier](https://github.com/47deg/metrifier), and got the following results when running locally (with benchmarks for HTTP and `Avro` for comparision - I haven't been able to get the proto benchmarks to run successfully 😞 )

```
Benchmark                          Mode  Cnt      Score     Error  Units
HttpBenchmark.createPerson        thrpt   40   9574.918 ± 226.283  ops/s
HttpBenchmark.getPerson           thrpt   40  12917.939 ± 269.448  ops/s
HttpBenchmark.getPersonLinks      thrpt   40   7040.148 ± 727.281  ops/s
HttpBenchmark.listPersons         thrpt   40  10567.644 ± 230.121  ops/s
HttpBenchmark.programComposition  thrpt   40   1320.074 ± 129.642  ops/s
```
```
Benchmark                             Mode  Cnt      Score     Error  Units
RPCAvroBenchmark.createPerson        thrpt   40  14577.077 ± 421.652  ops/s
RPCAvroBenchmark.getPerson           thrpt   40  15219.097 ± 607.676  ops/s
RPCAvroBenchmark.getPersonLinks      thrpt   40  11002.357 ± 533.516  ops/s
RPCAvroBenchmark.listPersons         thrpt   40  12688.747 ± 680.821  ops/s
RPCAvroBenchmark.programComposition  thrpt   40   1729.936 ±  40.921  ops/s
```
```
Benchmark                                       Mode  Cnt     Score     Error  Units
RPCAvroWithSchemaBenchmark.createPerson        thrpt   40  7084.775 ± 231.250  ops/s
RPCAvroWithSchemaBenchmark.getPerson           thrpt   40  8996.336 ± 254.989  ops/s
RPCAvroWithSchemaBenchmark.getPersonLinks      thrpt   40  7153.109 ± 287.379  ops/s
RPCAvroWithSchemaBenchmark.listPersons         thrpt   40  8342.474 ± 229.561  ops/s
RPCAvroWithSchemaBenchmark.programComposition  thrpt   40   995.148 ±  65.668  ops/s
```
As you can see, `AvroWithSchema` performs considerably worse than both `Avro` and HTTP. I know of at least one optimization that improves the `AvroWithSchema` throughput by quite a bit, and I will add that change soon. I think I know of a couple of other tweaks that could be made that would improve performance even more, but I would like to get feedback on the PR in general before spending more time on optimizations (I won't spend the time if the PR is to be rejected on principle 🙂 )

After the most recent changes to support scala source generation from `.avdl` files for the `Avro` `SerializationType`, I may need a little bit of help getting idl and source generation sorted out for this new `SerializationType`. 

Any feedback would be greatly appreciated! Thanks!


tyler-clark:
Apparently I might also need help fixing tests for this PR, as they all seem to be passing locally for me.

modules/server/src/test/scala/fs2/RPCTests.scala
```diff
@@ -80,7 +86,7 @@ class RPCTests extends RpcBaseTestSuite with BeforeAndAfterAll {
         .unsafeRunSync() shouldBe dResult33
     }
 
-    "be able to run client bidirectional streaming services" ignore {
+    "be able to run client bidirectional streaming services" in {
```
juanpedromoreno: This is related to a random failure. That's one of the reasons why your build failed: https://github.com/frees-io/freestyle-rpc/issues/164

modules/server/src/test/scala/fs2/RPCTests.scala
```diff
@@ -91,21 +97,37 @@ class RPCTests extends RpcBaseTestSuite with BeforeAndAfterAll {
...
+      freesRPCServiceClient
+        .biStreamingWithSchema(Stream.fromIterator[ConcurrentMonad, E](eList.iterator))
+        .compile
+        .toList
+        .unsafeRunSync()
+        .distinct shouldBe eList
+
+    }
+
+    "be able to run multiple rpc services" in {
```
juanpedromoreno: Same here

modules/server/src/test/scala/fs2/RPCTests.scala
tyler-clark: @juanpedromoreno thanks for pointing this out! I went ahead and re-ignored the `multiple rpc services` tests with https://github.com/frees-io/freestyle-rpc/pull/215/commits/2aaca90844031e1262690bca70341c15a7eaa28b. While the `bidirectional streaming services` tests were previously ignored, they don't seem to be failing currently. Would you like me to still set them to `ignore`, or are they ok to be left running?

L-Lavigne:
Great contribution @tyler-clark! Thanks for running the benchmarks as well.

As for the latest test failures (those in your new tests), I haven't been able to re-create the CI conditions to reproduce them. (We're also seeing a [gRPC error reporting issue](https://github.com/grpc/grpc-java/issues/1913) which complicates diagnosis).

In the short term we'll probably need to mark those tests as ignored, or preferably tag them as running only locally which I'm working on building support for (https://github.com/frees-io/freestyle-rpc/pull/216).

L-Lavigne:
Hey Tyler, #216 has been merged so if you re-merge `master` you can use the new `[ignoreOnTravis](https://github.com/frees-io/freestyle-rpc/pull/216/files#diff-375b404637b5ab345be83cfa0edfaed5R44)` method to exclude the failing tests from CI execution.

juanpedromoreno:
I think we can merge this now, and tag them as `ignore` in a separate PR. I'd like to do some other checks. Thanks @tyler-clark!

PR #223, opened 2018-04-05
Create example structure

AdrianRaFo:
Create an empty structure for the samples and the modules for the first one

### DIFF
--- a/modules/examples/routeguide/client/src/main/scala/Client.scala
+++ b/modules/examples/routeguide/client/src/main/scala/Client.scala
@@ -1,4 +1,4 @@
-package routeguide.client
+package example.routeguide.client
 
 
 class Client {
--- a/modules/examples/routeguide/client/src/main/scala/Client.scala
+++ b/modules/examples/routeguide/client/src/main/scala/Client.scala
@@ -1,0 +1,6 @@
+package routeguide.client
+
+
+class Client {
+
+}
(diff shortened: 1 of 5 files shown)


modules/examples/routeguide/client/src/main/scala/Client.scala
```diff
@@ -0,0 +1,6 @@
+package routeguide.client
```
eperinan: Let's looking for a different package naming for all the examples in order to have unified all examples with the same prefix.

For example, `package examples.routeguide.client``

What do you think?

modules/examples/routeguide/protocol/src/main/scala/Protocol.scala
```diff
@@ -0,0 +1,6 @@
+package routeguide.protocol
```
eperinan: Same as above...

modules/examples/routeguide/server/src/main/scala/Server.scala
```diff
@@ -0,0 +1,6 @@
+package routeguide.server
```
eperinan: Same as above...

modules/examples/routeguide/service/src/main/scala/Service.scala
```diff
@@ -0,0 +1,6 @@
+package routeguide.service
```
eperinan: Same as above...

Issue #220, opened 2018-04-05
Create the server module with the protocol


AdrianRaFo:
Create the server module with the protocol using IO

PR #232, opened 2018-04-10
Create example route guide server module

AdrianRaFo:
This pr adds the server module to the route guide example
The server module contains a server with Task and Freestyle-RPC
Fixes: #220

### DIFF
--- a/modules/examples/routeguide/server/src/main/scala/io/ServerIO.scala
+++ b/modules/examples/routeguide/server/src/main/scala/io/ServerIO.scala
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package example.routeguide.server.io
 
 import cats.effect.IO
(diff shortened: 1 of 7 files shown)


modules/examples/routeguide/server/src/main/scala/io/ServerIO.scala
```diff
@@ -0,0 +1,46 @@
...
+import cats.effect.IO
+import freestyle.rpc.server._
+import freestyle.rpc.server.config.BuildServerFromConfig
+import freestyle.rpc.server.{AddService, GrpcConfig, ServerW}
+import freestyle.tagless.config.implicits._
+import example.routeguide.server.handlers.RouteGuideServiceHandler
+import example.routeguide.protocol.Protocols.RouteGuideService
+import example.routeguide.runtime._
+
+object ServerIO {
```
juanpedromoreno: The nested trait and object can be pulled up from this object, so we could remove this object.

Issue #222, opened 2018-04-05
Move route guide example to RPC repository


AdrianRaFo:
- [x] Create the examples structure #217
 - [x] Add protocol sbt module to the example #218
 - [x] Add service sbt module to the example #219
 - [x] Create the server module with the protocol #220
 - [x] Create the client module with the protocol #221

PR #236, opened 2018-04-11
Add route guide example

AdrianRaFo:
Add the route guide example from `Freestyle-RPC-examples` to `Freestyle-RPC` under the new examples folder.
This example is composed of 5 modules:
 * Client: with the example clients
 * Server: with the example server
 * Protocol: with the example protocol
 * Runtime: with the example runtime base
 * Common: with the common dependencies of client and server module

Fixes: #222


modules/examples/routeguide/README.md
```diff
@@ -0,0 +1,44 @@
+
+# Routeguide-example
```
juanpedromoreno: We need to sync-up this file with the root README file

PR #241, opened 2018-04-16
Allow a sequence of source generated directories

AdrianRaFo:
This pr includes:
 - Moving the `srcGenSourceDir` setting from `File` to `Seq[File]` to allow get `avdl` or `proto` file from different directories
 - Create a new setting called `srcGenSourceFromJarsDir` to specify where extract the jars source files
 - Change the default value for `srcGenSourceDir` by the resources directory and the `srcGenSourceFromJarsDir` value


modules/idlgen/plugin/src/main/scala/IdlGenPlugin.scala
```diff
@@ -54,8 +54,12 @@ object IdlGenPlugin extends AutoPlugin {
           "Protobuf, Avro or AvroWithSchema are the current supported serialization types. " +
           "By default, the serialization type is 'Avro'.")
 
-    lazy val srcGenSourceDir: SettingKey[File] =
-      settingKey[File]("The IDL directory, where your IDL definitions are placed.")
+    lazy val srcGenSourceDir: SettingKey[Seq[File]] =
```
L-Lavigne: Let's pluralize the setting name and description.

modules/idlgen/plugin/src/main/scala/IdlGenPlugin.scala
```diff
@@ -79,8 +83,9 @@ object IdlGenPlugin extends AutoPlugin {
     idlGenSourceDir := (Compile / sourceDirectory).value,
     idlGenTargetDir := (Compile / resourceManaged).value,
     srcGenSerializationType := "Avro",
+    srcGenSourceFromJarsDir := idlGenTargetDir.value / idlType.value,
+    srcGenSourceDir := Seq((Compile / resourceDirectory).value, srcGenSourceFromJarsDir.value),
```
L-Lavigne: Would making this a collection require changing the corresponding `srcGen` arg? (currently `srcGenSourceDir.value.allPaths.get.toSet`)

AdrianRaFo:
Closed by #243

PR #243, opened 2018-04-17
Allow a sequence of source generated directories

### LINKED ISSUE #240: Allows concatenate different directories to source files
We should 
 - Create a new setting to specify where should be placed the source files extracted from the jars with the following sample as a proposal default value:

```scala
    srcGenSourceFromJarsDir := idlGenTargetDir.value / idlType.value
```

 - Move the `srcGenSourceDir` setting from `File` to `Seq[File]` in order to provide the possibility of having more than one directory with the source files, like the following:

```scala
srcGenSourceDir := Seq(
   (Compile / resourceDirectory).value, 
   srcGenSourceFromJarsDir.value)
```

AdrianRaFo:
This PR supersedes #241 
Resolves #240 

It includes:
 - Moving the `srcGenSourceDir` setting from `File` to `Seq[File]` to allow get `avdl` or `proto` file from different directories
 - Create a new setting called `srcGenSourceFromJarsDir` to specify where extract the jars source files
 - Change the default value for `srcGenSourceDir` by the resources directory and the `srcGenSourceFromJarsDir` value


modules/idlgen/plugin/src/main/scala/IdlGenPlugin.scala
```diff
@@ -54,8 +54,12 @@ object IdlGenPlugin extends AutoPlugin {
           "Protobuf, Avro or AvroWithSchema are the current supported serialization types. " +
           "By default, the serialization type is 'Avro'.")
 
-    lazy val srcGenSourceDir: SettingKey[File] =
-      settingKey[File]("The IDL directory, where your IDL definitions are placed.")
```
juanpedromoreno: Let's deprecated this settings before removing it

modules/idlgen/plugin/src/main/scala/IdlGenPlugin.scala
```diff
@@ -76,11 +85,13 @@ object IdlGenPlugin extends AutoPlugin {
 
   lazy val defaultSettings: Seq[Def.Setting[_]] = Seq(
     idlType := "(missing arg)",
+    srcGenSerializationType := "Avro",
     idlGenSourceDir := (Compile / sourceDirectory).value,
     idlGenTargetDir := (Compile / resourceManaged).value,
-    srcGenSerializationType := "Avro",
-    srcJarNames := Seq.empty,
+    srcGenSourceFromJarsDir := idlGenTargetDir.value / idlType.value,
```
juanpedromoreno: `idlGenTargetDir` would be unrelated since it's the opposite direction (source code to IDL files).. I think we might want to set it as `(Compile / resourceManaged).value`, even knowing that would be the same.

AdrianRaFo: with the `/  idlType.value` or without it?

juanpedromoreno: with is fine.

docs/src/main/tut/idl-generation.md
```diff
@@ -227,7 +227,9 @@ Just like `idlGen`, `srcGen` and `srcGenFromJars` has some configurable settings
 * **`idlType`**: the type of IDL to generate from, currently only `avro`.
 * **`srcGenSerializationType`**: the serialization type when generating Scala sources from the IDL definitions. `Protobuf`, `Avro` or `AvroWithSchema` are the current supported serialization types. By default, the serialization type is `Avro`.
 * **`srcJarNames`**: the list of jar names containing the IDL definitions that will be used at compilation time by `srcGenFromJars` to generate the Scala Sources. By default, this sequence is empty.
+* **`srcGenSourceFromJarsDir`**: The IDL from jars directory, where your IDL definitions extracted from jars are placed. By default is `(Compile / resourceManaged) / idlType`, typically `target/scala-2.12/src_managed/main/avro`
```
juanpedromoreno: > : the directory where the IDL files extracted from the different `jar` files will be placed.

What do you think?

docs/src/main/tut/idl-generation.md
```diff
@@ -227,7 +227,9 @@ Just like `idlGen`, `srcGen` and `srcGenFromJars` has some configurable settings
...
 * **`srcGenSourceDir`**: the IDL source base directory, where your IDL files are placed. By default: `Compile / resourceDirectory`, typically `src/main/resources/`.
+* **`srcGenSourceDirs`**: the sequence of IDL source base directories, where your IDL files are placed. By default contains `srcGenSourceDir` and `srcGenSourceFromJarsDir`. In the jar, it will maintain the packages structure you define.
```
juanpedromoreno: > : the list of directories where your IDL files are placed. By default, it contains `srcGenSourceDir.value` and `srcGenSourceFromJarsDir.value` folders.

docs/src/main/tut/idl-generation.md
```diff
@@ -227,7 +227,9 @@ Just like `idlGen`, `srcGen` and `srcGenFromJars` has some configurable settings
...
 * **`srcGenTargetDir`**: the Scala target base directory, where the `srcGen` task will write the Scala files in subdirectories/packages based on the namespaces of the IDL files. By default: `Compile / sourceManaged`, typically `target/scala-2.12/src_managed/main/`.
 * **`genOptions`**: additional options to add to the generated `@rpc` annotations, after the IDL type. Currently only supports `"Gzip"`.
```
juanpedromoreno: At this point is where I would mention:

> Note: regarding `srcGenSourceDirs`, all the directories configured as the source, will be distributed in the resulting _jar_ artifact preserving the same folder structure as in the source.

PR #245, opened 2018-04-17
[WIP] - Allows generating code from IDL in jars and submodules

fedefernandez:
This is the solution that I've found for generating code from IDL present in jar files and submodules and you're running an application (or generating a fat jar).

In the `srcJarNames` you need to specify not only the jars but also the submodule names that contain the IDL files.

Potentially, we'd need to change the setting name for something like `srcIdlNames` or something like that. This also filters the files copied to `srcGenSourceFromJarsDir`, so probably we wouldn't need the `mappings in (Compile, packageBin) ~= { _.filter(!_._1.getName.endsWith(".class")) }` filter anymore.

Thoughts?


fedefernandez:
Closed in favor of #246

PR #253, opened 2018-04-25
Tag RPC Service

JesusMtnez:
Manage Tags using RPC service / client.

### DIFF
--- a/modules/examples/todolist/client/src/main/scala/ClientApp.scala
+++ b/modules/examples/todolist/client/src/main/scala/ClientApp.scala
@@ -17,6 +17,7 @@
 package examples.todolist.client
 
 import cats.effect.IO
+import cats.syntax.apply._
 import examples.todolist.client.ClientProgram._
 import examples.todolist.client.implicits._
 import org.log4s._
@@ -31,8 +32,7 @@ object ClientApp {
 
     logger.info(s"${Thread.currentThread().getName} Starting client...")
 
-    pongProgram[IO].unsafeRunSync()
-    tagProgram[IO].unsafeRunSync()
+    (pongProgram[IO] *> tagProgram[IO]).unsafeRunSync()
 
     logger.info(s"${Thread.currentThread().getName} Closing client...")
 
--- a/modules/examples/todolist/client/src/main/scala/ClientApp.scala
+++ b/modules/examples/todolist/client/src/main/scala/ClientApp.scala
@@ -32,6 +32,7 @@ object ClientApp {
     logger.info(s"${Thread.currentThread().getName} Starting client...")
 
     pongProgram[IO].unsafeRunSync()
+    tagProgram[IO].unsafeRunSync()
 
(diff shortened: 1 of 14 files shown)


modules/examples/todolist/client/src/main/scala/ClientProgram.scala
```diff
@@ -17,18 +17,27 @@
...
-  def clientProgram[M[_]: Monad](implicit client: PingPongClient[M]): M[Unit] =
+  def pongProgram[M[_]: Monad](implicit client: PingPongClient[M]): M[Unit] =
     client.ping()
+
+  def tagProgram[M[_]: Monad](implicit tagClient: TagClient[M]): M[Unit] =
+    for {
+      _         <- tagClient.reset()
+      optionTag <- tagClient.insert(TagRequest("tag"))
+    } yield {
+      //println(s">>> $optionTag")
```
AdrianRaFo: :fire:

JesusMtnez: My bad! :disappointed:

project/ProjectPlugin.scala
```diff
@@ -23,7 +23,7 @@ object ProjectPlugin extends AutoPlugin {
       val avrohugger: String         = "1.0.0-RC4"
       val catsEffect: String         = "0.10.1"
       val circe: String              = "0.9.3"
-      val frees: String              = "0.8.0"
+      val frees: String              = "0.8.1-SNAPSHOT"
```
juanpedromoreno: This should be 0.8.0

JesusMtnez: Totally agree! It's better this way

project/ProjectPlugin.scala
```diff
@@ -164,6 +164,7 @@ object ProjectPlugin extends AutoPlugin {
 
     lazy val exampleTodolistCommonSettings: Seq[Def.Setting[_]] = Seq(
       libraryDependencies ++= Seq(
+        "io.frees" %% "frees-todolist-lib" % V.frees,
```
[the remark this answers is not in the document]
JesusMtnez: Done

modules/examples/todolist/client/src/main/scala/ClientApp.scala
```diff
@@ -31,7 +31,8 @@ object ClientApp {
 
     logger.info(s"${Thread.currentThread().getName} Starting client...")
 
-    clientProgram[IO].unsafeRunSync()
+    pongProgram[IO].unsafeRunSync()
+    tagProgram[IO].unsafeRunSync()
```
juanpedromoreno: You could run both programs with just one effect, for example:

```scala
import cats.syntax.apply._
(pongProgram[IO] *> tagProgram[IO]).unsafeRunSync()
```

JesusMtnez: Thanks!

PR #271, opened 2018-05-22
BigDecimal serialization in protobuf and avro

fedefernandez:
Provides `BigDecimal` serialization in protobuf and avro.

In the case of avro, I've created the `SchemaFor`, `FromValue`, and `ToValue` instances that will be used in the shapeless derivation of case classes in avro4s. 

Also, a `Marshaller` is provided in avro for serializing `BigDecimal`s as requests and/or responses because Avro doesn't provide the automatic derivation of `FromRecord` and `ToRecord` instances for other types than products.

### DIFF
--- a/modules/server/src/test/scala/protocol/RPCBigDecimalTests.scala
+++ b/modules/server/src/test/scala/protocol/RPCBigDecimalTests.scala
@@ -18,6 +18,7 @@ package freestyle.rpc
 package protocol
 
 import cats.Applicative
+import cats.syntax.applicative._
 import org.scalatest._
 import freestyle.rpc.common._
 import freestyle.rpc.testing.servers.withServerChannel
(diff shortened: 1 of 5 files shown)


modules/server/src/test/scala/protocol/RPCBigDecimalTests.scala
```diff
@@ -0,0 +1,183 @@
...
+
+      @rpc(Avro) def bigDecimalAvro(bd: BigDecimal): F[BigDecimal]
+      @rpc(Avro) def bigDecimalAvroWrapper(req: Request): F[Response]
+
+      @rpc(AvroWithSchema) def bigDecimalAvroWithSchema(bd: BigDecimal): F[BigDecimal]
+      @rpc(AvroWithSchema) def bigDecimalAvroWithSchemaWrapper(req: Request): F[Response]
+
+    }
+
+    class RPCServiceDefImpl[F[_]](implicit A: Applicative[F]) extends RPCServiceDef[F] {
```
juanpedromoreno: `F: Applicative[F]`?

fedefernandez: 👍

PR #274, opened 2018-05-24
Bump avrohugger to 1.0.0-RC9

JesusMtnez:
Updating to this version includes support for `decimal` logical type in avro using `BigDecimal`.


.sbtopts
```diff
@@ -0,0 +1,7 @@
+-J-Xmx5g
```
L-Lavigne: Interesting, I haven't used this file before.

JesusMtnez: I was having some trouble executing the tests due to a lack of memory in sbt JVM. I have imported it from another project.

Issue #269, opened 2018-05-17
Avro Marshallers for Date and Timestamp


juanpedromoreno:
It'd be handy to have `frees-rpc` providing optional marshallers for serializing and deserializing dates, with no reference to a particular time zone or time of day.

Avro specs: https://avro.apache.org/docs/current/spec.html#Date

Initially, we could support:

#### Date

> The `date` logical type represents a date within the calendar, with no reference to a particular time zone or time of day.

#### Timestamp (millisecond precision)

> The `timestamp-millis` logical type represents an instant on the global timeline, independent of a particular time zone or calendar, with a precision of one millisecond.

As we need to take into account `Java 1.7`, we could create two new different sbt modules (artifacts):

* `frees-rpc-marshallers-jodatime`: `Java 1.7` users could include this one to serialize/deserialize dates based on:
  * Avro `date`: it can be managed by [org.joda.time.LocalDate](http://joda-time.sourceforge.net/apidocs/org/joda/time/LocalDate.html).
  * Avro `timestamp-millis`: it can be managed by [org.joda.time.LocalDateTime](http://joda-time.sourceforge.net/apidocs/org/joda/time/LocalDateTime.html).
* `frees-rpc-marshallers-javatime`: `Java +1.8` users could include this one to serialize/deserialize dates based on 
  * Avro `date`: it can be managed by [java.time.LocalDate](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html).
  * Avro `timestamp-millis`: it can be managed by [java.time.LocalDateTime](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html).

These two new modules can be placed at `modules/marshallers/jodatime` and `modules/marshallers/javatime`, respectively.

Internally, as part of the serialization/deserialization process, we can use `Long` as intermediate bytes representation.

raulraja:
Doesn't avro4s already support `java.time.*`?

AdrianRaFo:
Only for `LocalDate`, `LocalDateTime`

AdrianRaFo:
Also, the avro main page specify that all the dates types supported are internally converted to an `Int|` or a `Long`

fedefernandez:
More information:
* `pbdirect` doesn't have support for LocalDate or LocalDateTime
* `avro4s` serializes/deserializes the dates in `String` format:
[...]

PR #276, opened 2018-05-25
Utility method for encode joda time instances

AntonioMateoGomez:
Add utility method for encoding joda.time instances.
Also adds utility for encoding from int and long to Array[Byte] in the internal module


modules/internal/src/test/scala/util/EncoderUtilTest.scala
```diff
@@ -0,0 +1,55 @@
...
+      }
+    }
+
+    "allow to convert long to and from byteArray" in {
+      check {
+        { n: Long =>
+          val value: Array[Byte] = EncoderUtil.longToByteArray(n)
+
+          EncoderUtil.byteArrayToLong(value) shouldBe n
+          EncoderUtil.byteArrayToLong(value) == n
```
fedefernandez: This was a mistake in my first PR, you could remove the `shouldBe` statement

modules/marshallers/jodatime/src/test/scala/util/JodaTimeUtilTest.scala
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    "allow to convert LocalDateTime to and from long" in {
+      import com.fortysevendeg.scalacheck.datetime.joda.granularity.seconds
+      check {
+        forAll(genDateTimeWithinRange(from, range)) { dt: DateTime =>
+          val date: LocalDateTime = dt.toDateTime(DateTimeZone.UTC).toLocalDateTime
+          val value: Long         = JodaTimeUtil.JodaLocalDatetimeToLong(date)
+
+          JodaTimeUtil.longToJodaLocalDateTime(value) shouldBe date
```
fedefernandez: Same here

modules/marshallers/jodatime/src/test/scala/util/JodaTimeUtilTest.scala
```diff
@@ -0,0 +1,64 @@
...
+  "JodaTimeUtil" should {
+
+    "allow to convert LocalDate to and from int" in {
+      import com.fortysevendeg.scalacheck.datetime.joda.granularity.days
+      check {
+        forAll(genDateTimeWithinRange(from, range)) { dt: DateTime =>
+          val date: LocalDate = dt.toLocalDate
+          val value: Int      = JodaTimeUtil.jodaLocalDateToInt(date)
+
+          JodaTimeUtil.intToJodaLocalDate(value) shouldBe date
```
fedefernandez: Remove this line, please

modules/internal/src/test/scala/util/EncoderUtilTest.scala
```diff
@@ -0,0 +1,55 @@
...
+class EncoderUtilTest extends WordSpec with Matchers with Checkers {
+
+  "EncoderUtil" should {
+
+    "allow to convert int to and from byteArray" in {
+      check {
+        { n: Int =>
+          val value: Array[Byte] = EncoderUtil.intToByteArray(n)
+
+          EncoderUtil.byteArrayToInt(value) shouldBe n
```
fedefernandez: Please, remove this line

modules/marshallers/jodatime/src/main/scala/util/JodaTimeUtil.scala
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package freestyle.rpc
+package jodatime
+package util
+
+import org.joda.time._
+
+object JodaTimeUtil {
+
```
fedefernandez: Could you please add a `val` here with the initial date? Something like:
```scala
private[this] val initialDate = DateTime.now(DateTimeZone.UTC).withMillis(0)
```

modules/internal/src/test/scala/util/JavaTimeUtilTests.scala
```diff
@@ -49,8 +49,9 @@ class JavaTimeUtilTests extends WordSpec with Matchers with Checkers {
       check {
         forAll(genDateTimeWithinRange(from, range)) { zdt: ZonedDateTime =>
           val date  = zdt.toLocalDate
-          val value = JavaTimeUtil.localDateToByteArray(date)
-          JavaTimeUtil.byteArrayToLocalDate(value) == date
```
fedefernandez: Please, remove this test. We've removed this method.

modules/internal/src/test/scala/util/JavaTimeUtilTests.scala
```diff
@@ -71,8 +72,8 @@ class JavaTimeUtilTests extends WordSpec with Matchers with Checkers {
       check {
         forAll(genDateTimeWithinRange(from, range)) { zdt: ZonedDateTime =>
           val date  = zdt.toLocalDateTime
-          val value = JavaTimeUtil.localDateTimeToByteArray(date)
-          JavaTimeUtil.byteArrayToLocalDateTime(value) == date
```
fedefernandez: Same as above, please remove this test

modules/marshallers/jodatime/src/main/scala/util/JodaTimeUtil.scala
```diff
@@ -0,0 +1,43 @@
...
+  private[this] val initialDate = DateTime.now(DateTimeZone.UTC).withMillis(0)
+
+  def jodaLocalDateToInt(value: LocalDate): Int =
+    Days
+      .daysBetween(
+        DateTime.now(DateTimeZone.UTC).withMillis(0),
```
fedefernandez: Replace this by the initalDate val

modules/marshallers/jodatime/src/main/scala/util/JodaTimeUtil.scala
```diff
@@ -0,0 +1,43 @@
...
+        value.toDateTimeAtStartOfDay(DateTimeZone.UTC))
+      .getDays
+
+  def intToJodaLocalDate(value: Int): LocalDate =
+    DateTime.now().withMillis(0).plusDays(value).toLocalDate
```
fedefernandez: Same as above

build.sbt
```diff
@@ -265,6 +265,21 @@ lazy val `example-todolist-client` = project
   .settings(exampleTodolistCommonSettings)
   .disablePlugins(ScriptedPlugin)
 
+/////////////////////
+//// MARSHALLERS ////
+/////////////////////
+
+lazy val `marshallers-jodatime` = project
```
L-Lavigne: Should this really be in its own module? Makes sense if our practice is do so when introducing new dependencies, otherwise I suggest integrating in an existing module.

build.sbt
```diff
@@ -265,6 +265,21 @@ lazy val `example-todolist-client` = project
...
+  .in(file("modules/marshallers/jodatime"))
+  .dependsOn(common % "compile->compile;test->test")
```
L-Lavigne: Do we need all those dependencies? I don't see them imported in the code. (Possibly more dependencies are needed for unit tests, but those should be only in test scopes and not exported).

build.sbt
L-Lavigne: In any case if making a new module, we should add it to the list in `quickstart.md`.

fedefernandez: I think it's a good idea to put it in it's own module in order to avoid adding more transitive dependencies. And yes, good catch, we should add it to the quickstart

build.sbt
fedefernandez: I think they will be used in the next PR for adding the marshallers but we could remove them for now 👍

build.sbt
L-Lavigne: Sounds good. We could also add a sentence or two in the docs about the new marshallers.

docs/src/main/tut/quickstart.md
```diff
@@ -28,6 +28,7 @@ It's divided into multiple and different artifacts, grouped by scope:
 `frees-rpc-client-netty` | Client | Yes* | Mandatory on the client side if we are using `Netty` in the server side.
 `frees-rpc-client-okhttp` | Client | Yes* | Mandatory on the client side if we are using `OkHttp` in the server side.
 `frees-rpc-config` | Server/Client | No | Provides configuration helpers using [frees-config] to load the application configuration values.
+`frees-rpc-marshallers` | Server/Client | No | Provides marshallers for serializing and deserializing the `LocalDate` and `LocalDateTime` joda instances.
```
L-Lavigne: `frees-rpc-marshallers-jodatime`

build.sbt
fedefernandez: I'd add the information in the docs once we solved #269 because now it's not ready for its usage.  @L-Lavigne @juanpedromoreno thoughts?

juanpedromoreno:
This is no longer blocked or waiting for the 0.14.0 ;)

modules/internal/src/test/scala/util/EncoderUtilTest.scala
```diff
@@ -0,0 +1,53 @@
...
+import org.scalacheck.Prop._
+import org.scalatest.prop.Checkers
+
+class EncoderUtilTest extends WordSpec with Matchers with Checkers {
+
+  "EncoderUtil" should {
+
+    "allow to convert int to and from byteArray" in {
+      check {
+        { n: Int =>
```
fedefernandez: here is missing the `forAll` right? maybe I'm missing something but I think this should be:
```scala
check {
  forAll { n: Int =>
    val value: Array[Byte] = EncoderUtil.intToByteArray(n)
    EncoderUtil.byteArrayToInt(value) == n
  }
}
```

AntonioMateoGomez: Yes Fede, you are right.

fedefernandez: Ok, we can tackle it in the next PR. Thanks

PR #287, opened 2018-06-01
Generate imports instead of using complete path

JesusMtnez:


juanpedromoreno:
@JesusMtnez What was the reason to label this as blocked? Thanks.

JesusMtnez:
As far as I remembered, this approach doesn't solve the issue it was supposed to solve (https://github.com/frees-io/freestyle-rpc/issues/288). That's why I marked as blocked.

fedefernandez:
Closing this because it's a bit outdated and doesn't solve the issue already fixed in the new avro4s versions.

PR #297, opened 2018-06-04
decouple from frees-async

pepegar:
bring the listenableFuture2Async function to the project, and don't
use frees-async-* packages

FIXES #297

### DIFF
--- a/modules/internal/src/main/scala/client/monixCalls.scala
+++ b/modules/internal/src/main/scala/client/monixCalls.scala
@@ -19,12 +19,13 @@ package internal
 package client
 
 import cats.effect.{Async, LiftIO}
-import freestyle.async.guava.implicits._
-import freestyle.async.catsEffect.implicits._
-import io.grpc.{CallOptions, Channel, MethodDescriptor}
+import com.google.common.util.concurrent._
 import io.grpc.stub.{ClientCalls, StreamObserver}
+import io.grpc.{CallOptions, Channel, MethodDescriptor}
+import java.util.concurrent.{Executor => JavaExecutor}
 import monix.execution.Scheduler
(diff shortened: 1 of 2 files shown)


modules/internal/src/main/scala/client/monixCalls.scala
```diff
@@ -79,4 +80,20 @@ object monixCalls {
             outputObserver
         ))
     )
+
+  private[this] def listenableFuture2Async[F[_], A](
+      fa: => ListenableFuture[A])(implicit F: Async[F], E: ExecutionContext): F[A] =
```
pepegar: I've changed teh `AsyncContext[F]` requirement with an `Async[F]` requirement, which is the same idea.

juanpedromoreno:
is this fixing #297 ?

modules/internal/src/main/scala/client/monixCalls.scala
juanpedromoreno: Maybe we can restrict the implicit req to the `Effect` typeclass?

pepegar:
Ah yes, I forgot mentioning it in the comment

modules/internal/src/main/scala/client/monixCalls.scala
pepegar: `Effect` is an `Async`, so it would be widening. Also, we're not using `runAsync` on the `F`, so we don't need `Effect`, right?

juanpedromoreno:
I'd suggest removing this: https://github.com/frees-io/freestyle-rpc/blob/master/project/ProjectPlugin.scala#L54.

We can use the grpc guava version, which it's already available in the classpath.

PR #300, opened 2018-06-05
Decouple frees config

### LINKED ISSUE #294: Decouple from frees-config
we're depending on frees-config for the config module (on which others depend). We need to stop depending on it for #290

Related to https://github.com/frees-io/freestyle-rpc/issues/290

pepegar:
In this PR I:

- create a bare minimum `ConfigM` tagless to be used wherever `freestyle.tagless.config` were used before
- remove a couple of `@module` annotations
- bump `frees-todolist-lib`to `v0.8.1`
- bump sbt-freestyle to `0.3.23`

FIXES #294

### DIFF
--- a/modules/config/src/main/scala/client/ChannelConfig.scala
+++ b/modules/config/src/main/scala/client/ChannelConfig.scala
@@ -18,27 +18,46 @@ package freestyle.rpc
 package client
 package config
 
-import cats.Functor
-import cats.syntax.either._
+import cats.effect.Sync
 import cats.syntax.functor._
-import freestyle.tagless._
-import freestyle.tagless.config.ConfigM
+import cats.syntax.flatMap._
+import cats.syntax.either._
 
-@module
-trait ChannelConfig {
+import freestyle.rpc.config.ConfigM
 
-  implicit val functor: Functor
+import com.typesafe.config.ConfigException.Missing
 
-  val configM: ConfigM
+trait ChannelConfig[F[_]] {
+
+  implicit def sync: Sync[F]
+  implicit def configM: ConfigM[F]
+
   val defaultHost: String = "localhost"
   val defaultPort: Int    = freestyle.rpc.server.defaultPort
 
-  def loadChannelAddress(hostPath: String, portPath: String): FS[ChannelForAddress] =
-    configM.load map (config =>
-      ChannelForAddress(
-        config.string(hostPath).getOrElse(defaultHost),
-        config.int(portPath).getOrElse(defaultPort)))
+  def loadChannelAddress(hostPath: String, portPath: String): F[ChannelForAddress] =
+    for {
+      config <- configM.load
+      host <- sync.pure(
+        Either
+          .catchOnly[Missing](config.getString(hostPath)))
+      port <- sync.pure(Either.catchOnly[Missing](config.getInt(portPath)))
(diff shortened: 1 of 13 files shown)


modules/config/src/main/scala/client/ChannelConfig.scala
```diff
@@ -18,27 +18,46 @@ package freestyle.rpc
...
+      target <- sync.pure(Either.catchOnly[Missing](config.getString(targetPath)))
+    } yield ChannelForTarget(target.getOrElse("target"))
+
+}
+
+object ChannelConfig {
+  def apply[F[_]](implicit S: Sync[F], C: ConfigM[F]): ChannelConfig[F] = new ChannelConfig[F] {
+    def sync    = S
+    def configM = C
+  }
```
juanpedromoreno: I'd suggest the following approach, more natural perhaps?

```scala
class ChannelConfig[F[_]](implicit S: Sync[F], C: ConfigM[F]) {

  val defaultHost: String = "localhost"
  val defaultPort: Int    = freestyle.rpc.server.defaultPort

  def loadChannelAddress(hostPath: String, portPath: String): F[ChannelForAddress] =
    for {
      config <- C.load
      host <- S.pure(
        Either
          .catchOnly[Missing](config.getString(hostPath)))
      port <- S.pure(Either.catchOnly[Missing](config.getInt(portPath)))
    } yield ChannelForAddress(host.getOrElse(defaultHost), port.getOrElse(defaultPort))

  def loadChannelTarget(targetPath: String): F[ChannelForTarget] =
    for {
      config <- C.load
      target <- S.pure(Either.catchOnly[Missing](config.getString(targetPath)))
    } yield ChannelForTarget(target.getOrElse("target"))

}

object ChannelConfig {
  def apply[F[_]](implicit S: Sync[F], C: ConfigM[F]): ChannelConfig[F] = new ChannelConfig[F]

  implicit def defaultChannelConfig[F[_]](implicit S: Sync[F], C: ConfigM[F]): ChannelConfig[F] =
    apply[F]
}
```

modules/config/src/main/scala/server/ServerConfig.scala
```diff
@@ -18,26 +18,44 @@ package freestyle.rpc
...
+
+import freestyle.rpc.config.ConfigM
+
+import com.typesafe.config.ConfigException.Missing
 
-@module
-trait ServerConfig {
+trait ServerConfig[F[_]] {
+
+  implicit def sync: Sync[F]
```
juanpedromoreno: Same here

modules/config/src/main/scala/config.scala
```diff
@@ -0,0 +1,33 @@
...
+
+import cats.effect.Sync
+import pureconfig._
+import com.typesafe.config.{Config, ConfigFactory}
+
+object config {
+
+  trait ConfigM[F[_]] {
+    def load: F[Config]
+  }
```
fedefernandez: I'm wondering if we could add a new method to this algebra:
```scala
def get[A](from: (Config) => A): F[Either[Missing, A]]
```
With the implementation:
```scala
def get[A](getValue: => A): F[Either[Missing, A]] =
  S.pure(Either.catchOnly[Missing](getValue))
```

pepegar: I think that, at some point in the near future, we'll reintroduce `frees-config`, and we could maybe add it there? this implementation is temporary. What do you think?

PR #307, opened 2018-06-07
WIP: bump fs2 reactive streams 0.6.0

pepegar:
# WIP: Do  not merge (waiting for monix to release a version depending on cats-effect 1.0.0-*)

- update fs-reactive-streams to 0.6.0
- migrate changes to adopt new fs2-rs (change `Effect` with `ConcurrentEffect`)
- bump monix to 3.0.0-RC1

---

There is a binary incompatibility in monix right now.  `monix:3.0.0-RC1` (current last version) depends on `cats-effect:0.10`, which is binary incompatible with `cats-effect:1.0.0-RC2` (the version `fs2-reactive-streams:0.6.0` depends on.

I don't think we can advance much more on this, In the beginning I thought that excluding monix' transitive dep on cats-effect would help but nope.

FIXES #303

PR #324, opened 2018-06-08
WIP: referentially transparent clients

### LINKED ISSUE #305: RPC Clients and Referential Transparency
As per some discussion some weeks ago with @fedefernandez and later on with @tbrown1979 about the RPC clients, the fact of instantiating a client is side-effect itself since we are trying to connect with the server, which is not helping to make the code referentially transparent. Therefore, we want to change our RPC client builders to support this idea.

Potentially we want to use `fs2.Stream.bracket` to be able to create RPC clients with this shape: `Stream[F, YourRPCClient[F]]`.

pepegar:
# WIP: Do not merge

Currently is not building because we're still depending on monix to release a cats-effect `1.0.0-RC2` compatible version. Also some more work needs to be done finishing the integration.

Waiting for monix/monix#681 to be released

---

This PR:
- bumps cats-effect to `1.0.0-RC2` (as #307 does)
- purifies the way we're instantiating GRPC's `ManagedChannel`s
- makes the instantiation of `Client`s more referentially transparent by bracketing over `F[Channel]` and `shutdown()` ing after use.
- seals `ManageChannelConfig`
- makes `NettyChannelConfig` not inherit `ManageChannelConfig` (it can't now)
- uses an `Either` for configuring netty clients (see [here](https://github.com/frees-io/freestyle-rpc/compare/ppg/305-ref-transparency-clients?expand=1#diff-38417b33d7ce2f25019cfcba711b5d0cR27))
- integrates those changes in the rest of the codebase

FIXES #305

### DIFF
--- a/modules/client-netty/src/main/scala/NettyChannelConfig.scala
+++ b/modules/client-netty/src/main/scala/NettyChannelConfig.scala
@@ -24,7 +24,7 @@ import io.grpc.netty.NegotiationType
 import io.netty.channel.{Channel, ChannelOption, EventLoopGroup}
 import io.netty.handler.ssl.SslContext
 
-sealed trait NettyChannelConfig                                         extends ManagedChannelConfig
+sealed trait NettyChannelConfig                                         extends Product with Serializable
 final case class NettyChannelType(channelType: Class[_ <: Channel])     extends NettyChannelConfig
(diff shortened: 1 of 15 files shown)


juanpedromoreno:
Superseded by #484

PR #325, opened 2018-06-12
implement Wart suppression

pepegar:
I've created a SupressWarts[A] typeclass with a single method
supressWarts(warts: _*)(a: A) to implement wart supression on several
different language constructs:

- methods (`DefDef`)
- values (`ValDef`)
- classes (`ClassDef`)

Since I'm creating instances of that typeclass for the above mentioned
types, i've left both the typeclass definition and the companion
object inside the def method, but the definition could potentially be
moved outside.

The downside of this implementation is that there's a lot of
boilerplate, but there's no homologue to the `mod""` interpolator for
scalamacros.  I've also thought of implementing the supression on a
structural type `T: {def mods: Modifiers}`, but I find it much more
ugly than current implementation.

### DIFF
--- a/modules/internal/src/main/scala/service.scala
+++ b/modules/internal/src/main/scala/service.scala
@@ -28,6 +28,68 @@ object serviceImpl {
     import c.universe._
     import Flag._
 
+    trait SupressWarts[T] {
+      def supressWarts(warts: String*)(t: T): T
+    }
+
+    object SupressWarts {
+      def apply[A](implicit A: SupressWarts[A]): SupressWarts[A] = A
+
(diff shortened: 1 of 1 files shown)


pepegar:
The typeclass we _could_ move it to another file, but the instances need to be inside the `service` method, because we're creating the instances for path dependent types (dependent on `c: blackbox.Context`).

This, and the fact that the typeclass is only used within the method, makes me think that's better leaving it there.  What do you think?

juanpedromoreno:
I see, didn't realize about it, so yes, it makes sense to me to leave it where it is now ;)

✅

PR #326, opened 2018-06-14
migrate idlgen to scalamacros

### LINKED ISSUE #298: Migrate IDL Generation
The `idlGen` sbt plugin leverages the macro-parsed `@service` definitions. As those get converted in #291, we'll need to migrate the plugin as well.

One possibility to explore is to _not_ use macros for the plugin, since it only reads the definitions and doesn't modify them or generate any code. Using some reflection-based API could be advantageous if it gives us more flexibility in handling types whose declarations deviate from the narrow range of textual statements we support (see https://github.com/frees-io/sbt-freestyle-protogen/issues/19 from the old plugin project).

pepegar:
Migrate `idlgen` project to scalamacros.  The biggest challenge is that we need to use path dependent types for all scala's AST types.

In this PR I:
- Introduced `monocle-core` as a dependency to `internal`, for the  `AstOptics`
- Introduced `scala-compiler` as a dependency to `internal`. We need it to be able to parse `*.scala` files.
- Changed the implementation of idlgen from to one based on `scalamacros`

N.B: ll AST types are dependent on the value `val Toolbox : ToolBox[reflect.runtime.universe.type]`, which is instantiated in the [`package.scala`](https://github.com/frees-io/freestyle-rpc/pull/326/files#diff-9db96018808db27cc00a6f891a9c4cb7).

Fixes #298


modules/idlgen/core/src/main/scala/ScalaParser.scala
```diff
@@ -14,43 +14,75 @@
...
-      defn  <- definitions
-      option <- defn.annotationsNamed("option")
+      defn             <- definitions
+      option           <- annotationsNamed("option").getAll(defn)
       Seq(name, value) <- option.withArgsNamed("name", "value")
     } yield RpcOption(name.toString.unquoted, value.toString) // keep value quoting as-is
 
-    val messages: Seq[RpcMessage] = definitions.filter(_.hasAnnotation("message")).map { defn =>
-      RpcMessage(defn.name, defn.params)
+    val messages: Seq[RpcMessage] = definitions.head.collect {
```
juanpedromoreno: `.head` would be unsafe :/

modules/idlgen/core/src/main/scala/ScalaParser.scala
```diff
@@ -14,43 +14,75 @@
...
+      case ast._CaseClassDef(mod) if hasAnnotation("message")(mod) => mod
+    } map { defn =>
+      RpcMessage(defn.name.toString, params.getOption(defn).get) // TODO: wat
```
juanpedromoreno: Could we address this `.get`

modules/idlgen/core/src/main/scala/ScalaParser.scala
```diff
@@ -14,43 +14,75 @@
...
+    }
+
+    def getRequestsFromService(defn: Tree): List[RpcRequest] = {
+      val rpcMethods = ast._AnnotatedDefDef("rpc")
+
+      defn.collect({ case rpcMethods(x) => x }).map { x =>
+        val serializationType = idlType.getAll(x).head
+        val name              = x.name.toString
+        val requestType       = firstParamForRpc.getOption(x).get
+        val responseType      = returnTypeAsString.getOption(x).get
```
juanpedromoreno: Same here, some `.head` and `.get` in this code fragment

modules/idlgen/core/src/main/scala/ScalaParser.scala
pepegar: yep, good catch! I had notes about fixing all those but forgot :)

modules/internal/src/main/scala/service.scala
```diff
@@ -48,6 +48,7 @@ object serviceImpl {
         require(params.length == 1, s"RPC call $name has more than one request parameter")
         RpcRequest(name, params.head.tpt, d.tpt, findAnnotation(d.mods, "rpc").get.children.tail)
```
juanpedromoreno: `.head` 🔥

juanpedromoreno: and `.get.children`

PR #327, opened 2018-06-21
Macro conversion

### LINKED ISSUE #291: Migrate Macro Annotations
As part of the https://github.com/frees-io/freestyle-rpc/issues/290 plan, we need to migrate the current Scalameta macros to the new generation of [Macro Annotations](https://docs.scala-lang.org/overviews/macros/annotations.html).

The current scalameta annotations are:

Annotation | Scope | Arguments | Description
--- | --- | --- | ---
@service | `Trait` | - | Tags the trait as an [RPC] service, in order to derive server and client code (macro expansion).
@rpc | `Method` | (`SerializationType`, `Compression`) | Indicates that the method is an RPC request. As `SerializationType` parameter value, `Protobuf` and `Avro` are the current supported serialization methods. As `Compression` parameter value, only `Gzip` is supported.
@stream | `Method` | `[S <: StreamingType]` | Indicates the  method's streaming type: server, client, and bidirectional. Hence, the `S` type parameter can be `ResponseStreaming`, `RequestStreaming`, or `BidirectionalStreaming`, respectively.
@message | `Case Class` | - | Tags the case class as an RPC message.
@option | `Object` | `(name: String, value: Any)` | Defines the equivalent headers in `.proto` files.
@outputPackage | `Object` | `(value: String)` | Defines the `package` declaration in `.proto` files, and the `namespace` tag in `.avpr` files.
@outputName  | `Object` | `(value: String)` | Defines the `protocol` tag in `.avpr` files; if missing the `idlGen` tool will use the source file name (without `.scala` extension) instead.

For all these annotations, the only one that is being actually expanded (in terms of expanded code) is **@service**. The rest of them can be considered as like "anchors"/"settings" used by the **@service** annotation.

We could consider doing this work creating smaller issues under the feature-branch `feature/scalamacros-migration`, for example.

pepegar:
FIXES #291


pepegar:
do you thing we shall rebase `release/0.14.0` on `master`?

pepegar:
Closing this one after the rebase of `release/0.14.0`, will reopen the PR

PR #328, opened 2018-06-21
Macro conversion wip

### LINKED ISSUE #291: Migrate Macro Annotations

pepegar:
Substitutes #327
Fixes #291


docs/src/main/tut/streaming.md
```diff
@@ -127,7 +124,6 @@ object service {
      * @return Stream of server responses.
      */
     @rpc(Protobuf)
-    @stream[ResponseStreaming.type]
```
diesalbla: What is happened to the `@stream` annotation? Is it no longer supported or no longer needed?

modules/idlgen/core/src/main/scala/ScalaParser.scala
```diff
@@ -14,43 +14,73 @@
...
+        }
+      } yield RpcRequest(serializationType, name, requestType, responseType, streamingType)
 
-    val services: Seq[RpcService] = definitions.filter(_.hasAnnotation("service")).map { defn =>
-      RpcService(defn.name, ServiceAlg(defn).requests.map(req =>
-        RpcRequest(req.serialization, req.name.value, req.requestType, req.responseType, req.streamingType)))
     }
-    // format: ON
+
+    val services: Seq[RpcService] = for {
```
diesalbla: I think you can merge the `for` and the `collect`, by placing the `yield` expression at the RHS of the match, with `defn` replaced by `mod`. 
```Scala
input.collect {
  case ast._ClassDef(mod) if hasAnnotation("service")(mod) =>
      RpcService(mod.name.toString, getRequestsFromService(mod))  
```

modules/idlgen/core/src/main/scala/avro/AvroIdlGenerator.scala
```diff
@@ -92,19 +104,24 @@ trait AvroIdlGenerator extends IdlGenerator {
...
-    case targ"Double"     => "double"
-    case targ"String"     => "string"
-    case targ"Seq[$t]"    => AvroArray(mappedType(t))
-    case targ"List[$t]"   => AvroArray(mappedType(t))
-    case targ"Array[$t]"  => AvroArray(mappedType(t))
-    case targ"Option[$t]" => AvroOption(mappedType(t))
-    case targ"Empty.type" => AvroEmpty
-    case _                => typeArg.toString
+  def mappedType(typeArg: Tree): AvroType = typeArg match {
+    case ast._Ident(Ident(TypeName("Boolean"))) => "boolean"
```
diesalbla: For this six cases, in which you have very similar patterns, I would use an [extractor object](https://docs.scala-lang.org/tour/extractor-objects.html), called `BasicTree`, so you may turn the first case into 
```Scala
case BaseType("Int") => "int"
```
This is even better in the five applied-tree cases, in which you could write.
```Scala
case SingleAppliedTree( "Seq", t) => AvroArray(mappedType(t))
```

modules/idlgen/core/src/main/scala/proto/ProtoIdlGenerator.scala
```diff
@@ -79,31 +83,35 @@ object ProtoIdlGenerator extends IdlGenerator {
...
-    case targ"Long"        => "int64"
-    case targ"Float"       => "float"
-    case targ"Double"      => "double"
-    case targ"String"      => "string"
-    case targ"Array[Byte]" => "bytes"
-    case targ"Option[$t]"  => mappedType(t)
-    case targ"List[$t]"    => s"repeated ${mappedType(t)}"
-    case _                 => typeArg.toString
+  private def mappedType(typeArg: Tree): String = typeArg match {
+    case ast._Ident(Ident(TypeName("Boolean"))) => "bool"
```
diesalbla: Here you can reuse the extractors from my previous comment.

modules/idlgen/core/src/main/scala/ScalaParser.scala
pepegar: You're right! good catch :)

modules/idlgen/core/src/main/scala/avro/AvroIdlGenerator.scala
pepegar: Thanks! I was definitely not happy with those humongous matches, but didn't thought of an extractor :)

PR #330, opened 2018-06-26
marshallers as implicit params

### LINKED ISSUE #278: Marshallers Required Implicitly
The current `@service` macro is requiring the Marshallers with the `implicitly` identifier, this is preventing to provide other marshallers externally of the macro, since the generated companion object is not able to find them.

https://github.com/frees-io/freestyle-rpc/blob/9f898c9/modules/internal/src/main/scala/service.scala#L205-L206

A potential solution would be changing the current macro, where instead of a `val`, we implement them as a `def`, requiring them as part of the method definition. Example:

```scala
q"""
       $wartSuppress
       def ${Pat.Var.Term(descriptorName)}(implicit RQM: _root_.io.grpc.MethodDescriptor.Marshaller[$requestType], RSM: _root_.io.grpc.MethodDescriptor.Marshaller[$responseType]): _root_.io.grpc.MethodDescriptor[$requestType, $responseType] = {
         $encodersImport
         _root_.io.grpc.MethodDescriptor
           .newBuilder(RQM, RSM)
           .setType(${utils.methodType(streamingType)})
           .setFullMethodName(
             _root_.io.grpc.MethodDescriptor.generateFullMethodName(${Lit.String(algName.value)}, ${Lit
      .String(name.value)}))
           .build()
       }
      """
```

pepegar:
This PR:

- makes the `@service` macro take the serialization type as parameter (and removes it from `@rpc` macro)
- makes the `$methodDescriptor` method on `@service` annotated traits take marshallers for request and response as parameters

FIXES #278

### DIFF
--- a/modules/common/src/main/scala/protocol.scala
+++ b/modules/common/src/main/scala/protocol.scala
@@ -33,8 +33,6 @@ sealed abstract class CompressionType extends Product with Serializable
 case object Identity                  extends CompressionType
 case object Gzip                      extends CompressionType
 
-@deprecated("Streaming type is now determined from request and response types", "0.14.0")
-class stream[S <: StreamingType]       extends StaticAnnotation
 class message                          extends StaticAnnotation
 class option(name: String, value: Any) extends StaticAnnotation
 class outputPackage(value: String)     extends StaticAnnotation


modules/idlgen/core/src/main/scala/avro/AvroSrcGenerator.scala
```diff
@@ -127,7 +127,8 @@ object AvroSrcGenerator extends SrcGenerator {
 
     val serviceLines =
       if (requestLines.isEmpty) Seq.empty
-      else Seq(s"@service trait ${protocol.getName}[F[_]] {", "") ++ requestLines :+ "}"
+      else
+        Seq(s"@service($serializationType) trait ${protocol.getName}[F[_]] {", "") ++ requestLines :+ "}"
```
diesalbla: The `F[_]` parameter here, is it generated as a fresh symbol?

modules/common/src/main/scala/protocol.scala
```diff
@@ -37,8 +33,7 @@ sealed abstract class CompressionType extends Product with Serializable
 case object Identity                  extends CompressionType
 case object Gzip                      extends CompressionType
 
-class rpc(val serializationType: SerializationType, val compressionType: CompressionType = Identity)
-    extends StaticAnnotation
+class rpc(val compressionType: CompressionType = Identity) extends StaticAnnotation
```
diesalbla: I think that we could just have the compression type as another annotation, e.g. `@gzip`, instead of an argument of `@rpc`.

modules/idlgen/core/src/main/scala/avro/AvroSrcGenerator.scala
pepegar: This `AvroSrcGenerator` is a code generator, that generates a `*.scala` file from an `*.av[dl|pr]`, it doesn't do macros.

pepegar:
Hm... I like the idea! It would also align the way `@services` are defined with what `@tagless` or `@free` do in Freestyle.

however, for that change I'd like some consensus (since there can be corner cases for that decision).

What do you think @fedefernandez @juanpedromoreno @L-Lavigne ?

juanpedromoreno:
Actually, with the code generation from protocols, we cannot use the gzip annotation given it is not part either `avro` nor `protocolbuff` standards, which it would be a limitation in my opinion.

So, I'm wondering if the gzip argument should be also part of the `@service` annotation, together with the `SerializationType`.

delvr:
I like this a lot, great work @pepegar . 👍 

@diesalbla I like the idea of getting rid of `@rpc` altogether; when you say "deferred methods" I guess you mean abstract?

As for `gzip`, I'd also be in favor of moving it up into `@service`, possibly as its own annotation, unless we have real-world use cases of some methods being compressed and others not. I could see it happening, but then again the user could just split that into 2 services.

L-Lavigne:
I like this a lot, great work @pepegar . 👍

@diesalbla I like the idea of getting rid of @rpc altogether; when you say "deferred methods" I guess you mean abstract?

As for gzip, I'd also be in favor of moving it up into @service, possibly as its own annotation, unless we have real-world use cases of some methods being compressed and others not. I could see it happening, but then again the user could just split that into 2 services.

pepegar:
Sounds good to me, @juanpedromoreno ! I think it makes much more sense to have it all in the `@service` annotation. I'll go do it tomorrow and update this PR!

modules/common/src/main/scala/protocol.scala
```diff
@@ -37,8 +33,6 @@ sealed abstract class CompressionType extends Product with Serializable
 case object Identity                  extends CompressionType
 case object Gzip                      extends CompressionType
 
-class rpc(val serializationType: SerializationType, val compressionType: CompressionType = Identity)
-    extends StaticAnnotation
 @deprecated("Streaming type is now determined from request and response types", "0.14.0")
```
juanpedromoreno: I think we could remove the `stream`, since we are no longer supporting the @rpc annotation

modules/server/src/test/scala/fs2/Utils.scala
```diff
@@ -27,51 +27,43 @@ object Utils extends CommonUtils {
...
+      def unaryCompressedWithSchema(a: A): F[C]
+      def biStreamingCompressedWithSchema(oe: Stream[F, E]): Stream[F, E]
     }
 
+    // this companion objects are here to make sure @service supports
+    // companion objects
+    object ProtoRPCService          {}
+    object AvroRPCService           {}
+    object AvroWithSchemaRPCService {}
+
```
juanpedromoreno: 👍

PR #333, opened 2018-07-02
Refactor GrpcServer

peterneyens:
- Creating a server now happens via methods on the `GrpcServer` companion object. Creating a server is now also an effect, so the return types of those methods is `F[GrpcServer[F]]`.
- `ServerW` is dropped to longer expose `io.grpc.Server`, the only way to interact with the underlying server is through `GrpcServer[F]`.
- The implementation of `GrpcServer` is changed to use `Sync#delay` so exceptions thrown are caught in `F`.


modules/server/src/main/scala/GrpcServer.scala
```diff
@@ -73,5 +77,51 @@ trait GrpcServer[F[_]] { self =>
 }
 
 object GrpcServer {
-  def apply[F[_]](implicit F: GrpcServer[F]): GrpcServer[F] = F
+
+  // def server[F[_]](S: GrpcServer[F])(implicit B: Bracket[F, Throwable]): F[Unit] =
+  //   B.bracket(S.start)(_ => S.awaitTermination)(_ => S.shutdown)
```
juanpedromoreno: Why is the alternative above commented?

peterneyens: We are currently still on cats-effect 0.10.x. 
The commented version is possible after we have upgraded to 1.x

modules/server/src/main/scala/netty.scala
```diff
@@ -45,10 +45,13 @@ object netty {
   final case class PermitKeepAliveTime(keepAliveTime: Long, timeUnit: TimeUnit) extends GrpcConfig
   final case class PermitKeepAliveWithoutCalls(permit: Boolean)                 extends GrpcConfig
 
-  final case class NettyServerConfigBuilder(initConfig: ChannelFor, configList: List[GrpcConfig]) {
+  // ?
```
fedefernandez: Not sure what this `?` means

modules/server/src/test/scala/server/RpcServerTestSuite.scala
```diff
@@ -125,10 +125,10 @@ trait RpcServerTestSuite extends RpcBaseTestSuite {
 
   object implicits extends Helpers with DummyData {
 
-    def grpcServerHandlerTests[F[_]](implicit F: Applicative[F]): GrpcServer[F] = {
+    def grpcServerHandlerTests[F[_]](implicit F: Sync[F]): GrpcServer[F] = {
       new GrpcServer[F] {
 
-        def start(): F[Server] = F.pure(serverMock.start())
+        def start(): F[Unit] = F.pure { serverMock.start(); () }
```
fedefernandez: Not a big deal, but could this be written as `F.pure(serverMock.start()).void`?

modules/server/src/test/scala/server/handlers/GrpcServerHandlerTests.scala
```diff
@@ -30,13 +30,13 @@ class GrpcServerHandlerTests extends RpcServerTestSuite {
...
 
-  val handler: GrpcServerHandler[ConcurrentMonad] = new GrpcServerHandler[ConcurrentMonad]
+  val handler = GrpcServerHandler[ConcurrentMonad]
 
   "GrpcServer.Handler" should {
 
     "allow to start a GrpcServer" in {
 
-      runK(handler.start, serverMock).unsafeRunSync() shouldBe serverCopyMock
+      runK(handler.start, serverMock).unsafeRunSync() shouldBe (())
```
fedefernandez: If we're not checking the result, maybe we could just check that everything works fine with something like:
`runK(handler.start, serverMock).attempt.unsafeRunSync().isRight shouldBe true`

modules/server/src/main/scala/netty.scala
peterneyens: I didn't really like these case class builders, and wanted to see if I could clean those up a little bit as well, but I forgot about them.

Added a new commit moving / inlining the code to helper methods in `GrpcServer`.

PR #334, opened 2018-07-02
Avro Schema Backward and Forward Compatibility

### LINKED ISSUE #331: Spike: Add some test cases for model evolutions
This ticket should provide some unit tests to cover the interoperability for old clients that keep pushing evolved services using legacy schemas. Some of the scenarios we want to cover:

- Add new non-optional field in the request
- Add new optional field in the request
- Add item to a coproduct in the request
- Remove item to a coproduct in the request
- Change type of an existing field in the request
- Rename field in the request
- Remove field in the request
- Add new non-optional field in the response
- Add new optional field in the response
- Add item to a coproduct in the response
- Remove item to a coproduct in the response
- Change type of an existing field in the response
- Rename field in the response
- Remove field in the response

rafaparadela:
This PR resolves #331 

It covers some tests cases related to the evolution of the requests/responses, which provide a general idea about how models can evolve once have been exposed and are being consumed by RPC clients.

The tests cover changes both in the request and response, and this is a summary of the conclusions I got after analyzing how Avro work:

### Cases:

- **A**: Add new non-optional field in the request
- **B**: Add new optional field in the request
- **C**: Add item to a coproduct in the request
- **D**: Remove item to a coproduct in the request
- **E**: Replace item in a coproduct in the request
- **F**: Change type of an existing field in the request
- **G**: Rename field in the request
- **H**: Remove field in the request
- **I**: Add new non-optional field in the response
- **J**: Add new optional field in the response
- **K**: Add item to a coproduct in the response
- **L**: Remove item to a coproduct in the response
- **M**: Replace item in a coproduct in the response
- **N**: Change type of an existing field in the response
- **O**: Rename field in the response
- **P**: Remove field in the response

### A: Add new non-optional field in the request
Request: Provide default value to the new field.

```scala
case class Request(a: String, b: Int)
case class NewRequest(a: String, b: Int, c: Boolean = true)
```

### B: Add new optional field in the request
Request: This is just another case of the previous scenario. The solutions could be to provide a default value, that presumably could be None/null.

### C: Add item to a coproduct in the request
Request: No action required

```scala
case class Request(a: Int :+: String :+: CNil)
case class NewRequest(a: Int :+: String :+: Boolean :+: CNil)
```

### D: Remove item to a coproduct in the request
Request: The only way to deal with this situation is considering this as a change of type, see next case.

```scala
case class Request(a: Int :+: String :+: CNil)
case class NewRequest(b: String :+: CNil = Coproduct[String :+: CNil](""))
```

### E: Replace item in a coproduct in the request
Request: If the type is just replaced, it will work while the passed value is one of the types in common between the previous coproduct and the new one.

```scala
case class Request(a: Int :+: String :+: CNil)
case class NewRequest(a: Int :+: Boolean :+: CNil)
```

It will work if the request is:
```scala
Request(a = Coproduct[Int :+: String :+: CNil](10))
```

and it will fail if the request is:
```scala
Request(a = Coproduct[Int :+: String :+: CNil]("Hi"))
```

### F: Change type of an existing field in the request
Request: It's is impossible to deal with a type swapping if we maintain the same name of the field. So the solution is considering this replacement as a combination of removing the old field/type, and adding a new field with the new type with a default value.

```scala
case class Request(a: String, b: Int)
case class NewRequest(a: String, c: Boolean = true)
```

### G: Rename field in the request
Request: Renaming a field is a particular case of creating a field with the new name and default value, and remove the old one, that will be ignored if old clients pass it.

```scala
case class Request(a: String, b: Int)
case class NewRequest(a: String, c: Int = 0)
```

### H: Remove field in the request
Request: No action required. But note that the value will be ignored when old clients include it in the request.

```scala
case class Request(a: String, b: Int)
case class NewRequest(a: String)
```

### I: Add new non-optional field in the response
Response: No action required. Old clients will ignore the value of the new field.

```scala
case class Response(a: String, b: Int)
case class NewResponse(a: String, b: Int, c: Boolean)
```

### J: Add new optional field in the response
Response: This is a just a particular case of the previous scenario.

### K: Add item to a coproduct in the response
Response: Obviously, it fails when the value is the new item. So the solution is providing a default value to the old coproduct and creating a new field with the new coproduct.

```scala
case class Response(a: Int :+: String :+: CNil)
case class NewResponse(
      a: Int :+: String :+: CNil = Coproduct[Int :+: String :+: CNil](0),
      b: Int :+: String :+: Boolean :+: CNil)
```

### L: Remove item to a coproduct in the response
Response: No action required.

```scala
case class Response(a: Int :+: String :+: CNil)
case class NewResponse(a: Int :+: CNil)
```

### M: Replace item in a coproduct in the response
Response: As long as the value of the coproduct belongs to the previous version, the old client should be able to accept the response as valid.

```scala
case class Response(a: Int :+: String :+: CNil)
case class NewResponse(a: Int :+: Boolean :+: CNil)
```

### N: Change type of an existing field in the response
Response: It requires to provide a default value to the previous type, and to create a new field with the new type.

```scala
case class Response(a: String, b: Int)
case class NewResponse(a: String, b: Int = 123, c: Boolean)
```

### O: Rename field in the response
Response: It requires to remain the previous name/type with a default value and to add a new field with the same type.

```scala
case class Response(a: String, b: Int)
case class NewResponse(a: String, b: Int = 123, c: Int)
```

### P: Remove field in the response
Response: This evolution should never happen. Only under the special case that the old response has a default value for the field we want to drop, then this operation is feasible, just removing the field in the new version of the response.

```scala
case class Response(a: String, b: Int = 123)
case class NewResponse(a: String)
```


modules/server/src/test/scala/avro/RPCTests.scala
```diff
@@ -17,82 +17,228 @@
 package freestyle.rpc
 package avro
 
-import cats.Apply
-import freestyle.rpc.common._
+import freestyle.rpc.common.{ConcurrentMonad, _}
```
juanpedromoreno: `import freestyle.rpc.common._`

modules/server/src/test/scala/avro/RPCTests.scala
```diff
@@ -17,82 +17,228 @@
...
 
-      withServerChannel(ssd) { sc =>
-        val rpcServiceClient: service.RPCService.Client[ConcurrentMonad] =
-          service.RPCService.clientFromChannel[ConcurrentMonad](sc.channel)
+    "add a new non-optional field, and" should {
+      "be able to respond to an outdated request without the new value" in {
+        runTestProgram(serviceRequestAddedBoolean.RPCService.bindService[ConcurrentMonad], response) {
+          client =>
+            client.get(request).unsafeRunSync()
+        }
```
juanpedromoreno:
```scala
        runTestProgram(
          serviceRequestAddedBoolean.RPCService.bindService[ConcurrentMonad],
          response
        )(_.get(request).unsafeRunSync())
```

?

modules/server/src/test/scala/avro/RPCTests.scala
```diff
@@ -17,82 +17,228 @@
...
+      }
+      "be able to respond to an outdated request without the new value within a coproduct" in {
+        runTestProgram(
+          serviceRequestAddedBoolean.RPCService.bindService[ConcurrentMonad],
+          responseCoproduct(response)) { client =>
+          client.getCoproduct(requestCoproduct(request)).unsafeRunSync()
+        }
```
juanpedromoreno: Same suggestion here:

```scala
        runTestProgram(
          serviceRequestAddedBoolean.RPCService.bindService[ConcurrentMonad],
          responseCoproduct(response))(
          _.getCoproduct(requestCoproduct(request)).unsafeRunSync()
        )
```

modules/server/src/test/scala/avro/RPCTests.scala
rafaparadela: Done at https://github.com/frees-io/freestyle-rpc/pull/334/commits/60a5df470c0126baf3e98174aa976b95c0aa62c1

modules/server/src/test/scala/avro/RPCTests.scala
rafaparadela: Done at https://github.com/frees-io/freestyle-rpc/pull/334/commits/60a5df470c0126baf3e98174aa976b95c0aa62c1

modules/server/src/test/scala/avro/RPCTests.scala
rafaparadela: Done at https://github.com/frees-io/freestyle-rpc/pull/334/commits/74fc9fa975138ad02673aba2c68e0bdfcb1e499c

rafaparadela:
@juanpedromoreno @diesalbla I have added a new scenario

### E: Replace item in a coproduct in the request
Request: If the type is just replaced, it will work while the passed value is one of the types in common between the previous coproduct and the new one.

```scala
case class Request(a: Int :+: String :+: CNil)
case class NewRequest(a: Int :+: Boolean :+: CNil)
```

It will work if the request is:
```scala
Request(a = Coproduct[Int :+: String :+: CNil](10))
```

and it will fail if the request is:
```scala
Request(a = Coproduct[Int :+: String :+: CNil]("Hi"))
```

I added a couple of tests: 
- One to prove that it works when the value of the coproduct still exists in the new coproduct.
- One to prove that it fails when the value of the coproduct is no longer part of the coproduct.

rafaparadela:
Should I merge this @juanpedromoreno ?

juanpedromoreno:
Perhaps @diesalbla wants to add his feedback on this :)

fedefernandez:
@rafaparadela any reason for not adding the new case `E` applied to the `response`?

rafaparadela:
@fedefernandez Added the case you were missing at https://github.com/frees-io/freestyle-rpc/pull/334/commits/b47a0e0cfe44b9956ed71bcae91cc7f979e11eef

PR #337, opened 2018-07-09
Add client chache sub project

peterneyens:
Add a cache for RPC clients.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -73,6 +73,12 @@ lazy val `client-okhttp` = project
   .settings(clientOkHttpSettings)
   .disablePlugins(ScriptedPlugin)
 
+lazy val `client-cache` = project
+  .in(file("modules/client-cache"))
+  .settings(moduleName := "frees-rpc-client-cache")
+  .settings(clientCacheSettings)
+  .disablePlugins(ScriptedPlugin)
+
 ////////////////
 //// SERVER ////
 ////////////////
@@ -273,6 +279,7 @@ lazy val allModules: Seq[ProjectReference] = Seq(
   common,
   internal,
   client,
+  `client-cache`,
   `client-netty`,
   `client-okhttp`,
   server,
--- a/modules/client-cache/src/main/scala/ClientCache.scala
+++ b/modules/client-cache/src/main/scala/ClientCache.scala
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2017-2018 47 Degrees, LLC. <http://www.47deg.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package freestyle.rpc.client.cache
+
+import cats.effect.{Effect, Timer}
+import cats.instances.list._
+import cats.instances.tuple._
+import cats.syntax.apply._
+import cats.syntax.bifunctor._
+import cats.syntax.flatMap._
+import cats.syntax.foldable._
+import cats.syntax.functor._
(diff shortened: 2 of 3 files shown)


build.sbt
```diff
@@ -73,6 +73,13 @@ lazy val `client-okhttp` = project
   .settings(clientOkHttpSettings)
   .disablePlugins(ScriptedPlugin)
 
+lazy val `client-cache` = project
+  .in(file("modules/client-cache"))
+  .settings(moduleName := "frees-rpc-client-cache")
+  .settings(clientCacheSettings)
+  .settings(addCompilerPlugin(%%("paradise") cross CrossVersion.full))
```
peterneyens: Without this line I get a `Missing required plugin: macroparadise` compiler error.

project/ProjectPlugin.scala
```diff
@@ -79,6 +79,15 @@ object ProjectPlugin extends AutoPlugin {
       )
     )
 
+    lazy val clientCacheSettings: Seq[Def.Setting[_]] = Seq(
+      libraryDependencies := Seq(
+        %%("log4s", V.log4s),
+        %%("fs2-core", "0.10.1"),
```
peterneyens: This it the fs2 version used by the fs2-reactive-streams version that we are currently using.
This one is sadly still on cats-effect 0.8.0

build.sbt
```diff
@@ -73,6 +73,13 @@ lazy val `client-okhttp` = project
   .settings(clientOkHttpSettings)
   .disablePlugins(ScriptedPlugin)
 
+lazy val `client-cache` = project
```
juanpedromoreno: Let's add it to the "modules registry" below in this file

project/ProjectPlugin.scala
juanpedromoreno: I'd add the `0.10.1` fs2 val with the rest, same for the `better-monadic-for`

build.sbt
juanpedromoreno: It's weird, it should be provided by the project plugin: https://github.com/frees-io/freestyle-rpc/blob/14671779136aa05ea146ae3126ef410cbd4f52db/project/ProjectPlugin.scala#L209

modules/client-cache/src/main/scala/ClientCache.scala
```diff
@@ -0,0 +1,108 @@
...
+      timer.clockRealTime(MILLISECONDS).map(_.millis)
+
+    def create(ref: async.Ref[F, State]) = new ClientCache[Client, F] {
+      val getClient: F[Client[F]] = for {
+        hostAndPort <- getHostAndPort
+        now         <- nowUnix
+        (map, _)    <- ref.get
+        client <- map
+          .get(hostAndPort)
+          .fold {
```
diesalbla: Generally, I prefer using `fold`. However, once once each case starts spawning over three-four lines, I find it more readable to use a `match`: 
```Scala
client <- map.get(hostAndPort) match {
  case None =>  
      /** 6 lines of code*/
  case Some(clientMeta) =>
     /** 5 lines of code*/
} 
```

modules/client-cache/src/main/scala/ClientCache.scala
```diff
@@ -0,0 +1,108 @@
...
+                    .modify(_.leftMap(_ + (hostAndPort -> ClientMeta(client, close, now))))
+                    .as(client)
+            }
+          }(
+            clientMeta =>
+              F.delay(logger.debug(s"Reuse existing RPC client for $hostAndPort")) *>
+                ref
+                  .modify(_.leftMap(_.updated(hostAndPort, clientMeta.copy(lastAccessed = now))))
+                  .as(clientMeta.client))
+        (_, lastClean) <- ref.get
```
diesalbla: Can we extract lines 82-85 as a separate `def cleanIfStale(atTime: Instant): F[Unit]` method?

build.sbt
peterneyens: I was overriding the `libraryDependencies` before, but it looks like we only get the paradise plugin and a scalatest test dependency.
I switched back to `++=`, so should be fine now.

project/ProjectPlugin.scala
peterneyens: Updated to 0.10.5, also added an explicit dependency to fs2 0.10.5 to `internal` so that the transitive fs2 0.10.1 gets evicted.

juanpedromoreno: Cool, thanks @peterneyens

PR #341, opened 2018-07-16
Marshallers for serializing and deserializing joda.time dates

AntonioMateoGomez:
-Adds marshallers for serializing and deserializing joda.time dates
-Also fixes the Scalacheck test in EncoderUtilTest file, adding "forAll".


fedefernandez:
This is ready for review

PR #368, opened 2018-07-20
WIP: derive http client from service

pepegar:
This is a work in progress to track the status of HTTP client derivation from service protocol (protocol = service annotated trait).  The actual implementation:

- [x] creates a new `http` object within the  generated companion
- [x] puts a `HttpClient[F[_]: Sync]` inside the `http` object
- [x] creates a new method on `HttpClient` for each `@http` annotated method in the protocol, with all the needed.
- [ ] Derive HTTP from [Protobuf|Avro|AvroWithSchema] (like we're doing for RPC).

Some assumptions/questions I've taken while implementing this:

- all HTTP methods should be unary, meaning that they only take a parameter
- how do we treat headers? If they're a static value, we can pass them as params to the `@http` annotation, to try to leave the method as similar to the RPC definition as possible. Otherwise, i haven't found a solution for that, since it'll make method incompatible with RPC...

## Test

To see generated code, I've used this file

```scala
import freestyle.rpc._
import freestyle.rpc.internal.encoders.pbd._

@service(Avro) trait GithubService[F[_]] {

  @http(POST, "https://github.com/frees-io/freestyle-rpc/issues")
  def postIssue(id: Int): F[String]

  @http(POST, "https://github.com/frees-io/freestyle-rpc/pr")
  def postPullRequest(id: Int): F[String]

}
```

And loaded directly in the REPL:

```bash
$ sbt client/console
scala> :load test.scala
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -174,14 +174,14 @@ lazy val `idlgen-sbt` = project
   .settings(buildInfoKeys := Seq[BuildInfoKey](name, version, scalaVersion, sbtVersion))
   .settings(buildInfoPackage := "freestyle.rpc.idlgen")
 
-lazy val `http-server` =  project
-  .in(file("modules/http/server"))
+lazy val http =  project
+  .in(file("modules/http"))
   .dependsOn(common % "compile->compile;test->test")
   .dependsOn(internal)
-  .dependsOn(client % "test->test")
+  .dependsOn(client)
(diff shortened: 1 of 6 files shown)


L-Lavigne:
Closing this PR as resolved in #552. Thanks @pepegar and @rafaparadela !

PR #374, opened 2018-07-27
Customize the codecs used in services through sbt

### LINKED ISSUE #371: New Sbt code generation setting
As part of this ticket, we need to add a new sbt setting to be able to specify custom imports including custom marshallers for serializing/deserializing data.

The @fedefernandez's proposal would be:

(Extensible) ADT (provided by frees-rpc atm):

```scala
abstract class MarshallersImport(val marshallersImport: String) extends Product with Serializable
case object BigDecimalAvroMarshallers extends MarshallersImport("import ....")
case object JavaTimeDateAvroMarshallers extends MarshallersImport("import ....")
case object JodaDateTimeAvroMarshallers extends MarshallersImport("import ....")
```

New Sbt Setting:

```scala
lazy val idlGenCodecImports: SettingKey[List[MarshallersImport]]
```

Default value:

```scala
List(BigDecimalAvroMarshallers, JavaTimeDateAvroMarshallers)
```

This new setting would bring the ability to define the custom user marshallers in the `build.sbt`.

* [ ] Sbt Settings.
* [ ] Docs.

Related to #370

### LINKED ISSUE #372: Custom Marshallers in Code Generation
Depends on https://github.com/frees-io/freestyle-rpc/issues/371

We need to modify the current sbt-idl-gen-plugin to be able to do the code generation based on the new sbt setting added as part of #371.

* [ ] Code generation update.
* [ ] Scripted tests.

fedefernandez:
Fixes #371 
Fixes #372 

Currently, we've got both *generators* (`avro` and `idl`) under the same "`trait`" and because the imports are something specific of the `AvroSrcGenerator` I've decided to pass the new property `idlGenCodecImports` through the constructor instead of through the task params like [the options](https://github.com/frees-io/freestyle-rpc/compare/ff-371-sbt-codec-imports-generation?expand=1#diff-78d1790b16820366c416d210b1eb3015R158). Please, let me know if that makes sense.


modules/idlgen/core/src/main/scala/Model.scala
```diff
@@ -63,4 +63,16 @@ object Model {
...
+      extends Product
+      with Serializable
+
+  final case class CustomMarshallersImport(mi: String) extends MarshallersImport(mi)
+  case object BigDecimalAvroMarshallers
+      extends MarshallersImport("freestyle.rpc.internal.encoders.avro.bigdecimal._")
+  case object JavaTimeDateAvroMarshallers
+      extends MarshallersImport("freestyle.rpc.internal.encoders.avro.javatime._")
+  case object JodaDateTimeAvroMarshallers
+      extends MarshallersImport("freestyle.rpc.marshallers.jodaTimeEncoders.avro._")
```
juanpedromoreno: Should we add here the proto imports?

modules/idlgen/core/src/main/scala/avro/AvroSrcGenerator.scala
```diff
@@ -105,11 +109,10 @@ object AvroSrcGenerator extends SrcGenerator {
...
 
-    val importLines = Seq(
-      "import freestyle.rpc.internal.encoders.avro.bigdecimal._",
-      "import freestyle.rpc.internal.encoders.avro.javatime._",
-      "import freestyle.rpc.protocol._"
-    )
+    val importLines =
+      ("import freestyle.rpc.protocol._" :: marshallersImports
+        .map(_.marshallersImport)
+        .map("import " + _)).sorted
```
juanpedromoreno: Should we check if the user has specified the `import` reserved word before adding it?

modules/idlgen/plugin/src/main/scala/IdlGenPlugin.scala
```diff
@@ -108,7 +114,8 @@ object IdlGenPlugin extends AutoPlugin {
     srcGenSourceDirs := Seq(srcGenSourceDir.value),
     srcGenIDLTargetDir := (Compile / resourceManaged).value / idlType.value,
     srcGenTargetDir := (Compile / sourceManaged).value,
-    genOptions := Seq.empty
+    genOptions := Seq.empty,
+    idlGenCodecImports := List(BigDecimalAvroMarshallers, JavaTimeDateAvroMarshallers)
```
juanpedromoreno: Are we tying the default values to Avro? Do you think we could the `idlType` setting to figure out these values?

modules/idlgen/plugin/src/sbt-test/sbt-frees-rpc-idlgen/importMarshallers/project/plugins.sbt
```diff
@@ -0,0 +1 @@
+addSbtPlugin("io.frees" %% "sbt-frees-rpc-idlgen" % "0.14.2-SNAPSHOT")
```
juanpedromoreno: version should be `sys.props("version")`

modules/idlgen/core/src/main/scala/Model.scala
```diff
@@ -63,4 +63,16 @@ object Model {
       case _ => false
     }
   }
+
+  sealed abstract class MarshallersImport(val marshallersImport: String)
```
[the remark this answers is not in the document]
fedefernandez: There is a custom one for that:
https://github.com/frees-io/freestyle-rpc/pull/374/files#diff-9adb97b8bcc89ae7f61c2758ce7c7583R17

modules/idlgen/core/src/main/scala/Model.scala
fedefernandez: They are not supported, but we should, yeah

docs/src/main/tut/idl-generation.md
```diff
@@ -221,6 +221,8 @@ Just like `idlGen`, `srcGen` and `srcGenFromJars` has some configurable settings
   * All the source folders specified in the `srcGenSourceDirs` setting.
 * **`srcGenTargetDir`**: the Scala target base directory, where the `srcGen` task will write the Scala files in subdirectories/packages based on the namespaces of the IDL files. By default: `Compile / sourceManaged`, typically `target/scala-2.12/src_managed/main/`.
 * **`genOptions`**: additional options to add to the generated `@service` annotations, after the IDL type. Currently only supports `"Gzip"`.
+* **`idlGenCodecImports`**: additional imports to add on top to the generated service files. This property can be used for importing extra encoders/decoders for your services. By default: `List(BigDecimalAvroMarshallers, JavaTimeDateAvroMarshallers)`. The `JodaDateTimeAvroMarshallers` is also available, but it needs the dependency `frees-rpc-marshallers-jodatime`. You can also specify custom imports with the following:
```
juanpedromoreno: Maybe we can rename this to `idlGenMarshallerImports`, to be fully aligned, thoughts?

modules/idlgen/plugin/src/main/scala/IdlGenPlugin.scala
```diff
@@ -90,6 +91,11 @@ object IdlGenPlugin extends AutoPlugin {
     lazy val genOptions: SettingKey[Seq[String]] =
       settingKey[Seq[String]](
         "Options for the generator, such as additional @service annotation parameters in srcGen.")
+
+    lazy val idlGenCodecImports: SettingKey[List[MarshallersImport]] =
```
juanpedromoreno: `idlGenMarshallerImports`?

docs/src/main/tut/idl-generation.md
fedefernandez: Agree

modules/idlgen/core/src/test/scala/AvroScalaGeneratorArbitrary.scala
```diff
@@ -0,0 +1,105 @@
...
+         |
+         |  def sayHelloAvro(arg: foo.bar.HelloRequest): F[foo.bar.HelloResponse]
+         |
+         |  def sayNothingAvro(arg: Empty.type): F[Empty.type]
+         |
+         |}""".stripMargin.split("\n").filter(_.length > 0).toList
+  }
+
+  val importSliceGen: Gen[String] =
+    Gen.choose(4, 10).flatMap(Gen.listOfN(_, Gen.choose('a', 'z')).map(_.mkString("")))
```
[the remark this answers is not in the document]
fedefernandez: I want to keep the generated string bounded by some logic limits, I'll use the `alphaLowerChar` 👍

PR #375, opened 2018-07-30
Bumps io.grpc dependency

fedefernandez:
I tried to update the `netty-tcnative-boringssl-static` dependency, but it gave me an abstract method error execption.


juanpedromoreno:
Which version did you try?

Based on the table here: https://github.com/grpc/grpc-java/blob/master/SECURITY.md#netty

You need to bump it up to `2.0.8.Final`.

fedefernandez:
Cool, I tried 2.0.12. I'll try 2.0.8 here. Thanks

PR #378, opened 2018-08-07
document mandatory compiler plugin

maaaikoool:
I belive this should be documented, same as in https://github.com/frees-io/freestyle/blob/master/README.md#freestyle-artifacts

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,12 @@ It's divided into multiple and different artifacts, grouped by scope:
 * `Yes*`: on the client-side, you must choose either `Netty` or `OkHttp` as the transport layer.
 * `Provided*`: you don't need to add it to your build, it'll be transitively provided when using other dependencies.
 
+To use the project, add the following to your build.sbt:
+
+```scala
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,6 +45,12 @@ It's divided into multiple and different artifacts, grouped by scope:
 * `Yes*`: on the client-side, you must choose either `Netty` or `OkHttp` as the transport layer.
 * `Provided*`: you don't need to add it to your build, it'll be transitively provided when using other dependencies.
 
+To use the project, add the following to your build.sbt:
+
+```scala
+addCompilerPlugin("org.scalameta" % "paradise" % "3.0.0-M11" cross CrossVersion.full)
```
juanpedromoreno: Thanks for your time @maaaikoool , but the scalameta paradise plugin is no longer needed since `0.14.x`.

However, it would be handy to add what's really needed:

```scala
compilerPlugin("org.scalamacros" % "paradise" % "2.1.1" cross CrossVersion.patch)
```

maaaikoool: Fixed, many thanks!

PR #385, opened 2018-09-24
Allows Custom namespace for server/client metrics

juanpedromoreno:
This PR allows providing externally the namespace for the registered metrics. Example:

```scala
val configList: List[ManagedChannelConfig] =
        AddInterceptor(
          MonitoringClientInterceptor(
            Configuration
              .defaultAllMetrics
              .withCollectorRegistry(CR)
              .withNamespace("custom_namespace")
          )
        )
```

`.withNamespace("custom_namespace")` is the combinator added here. If it's not provided, by default will be `grpc`, the same used until now.


AdrianRaFo:
@juanpedromoreno it would be possible get the project name and add it to the default namespace?

juanpedromoreno:
Frees-rpc users could do it that way.

PR #386, opened 2018-09-25
Depending on Execution Context instead of Monix Scheduler

juanpedromoreno:
This PR replaces, where possible, the `monix.execution.Scheduler` dependency by `scala.concurrent.ExecutionContext`. Therefore, _RPC clients_ (based on RPC protocols) that are not depending on `Monix` in its classpath can interact with the RPC services without needing to add it.

### DIFF
--- a/modules/examples/todolist/runtime/src/main/scala/CommonRuntime.scala
+++ b/modules/examples/todolist/runtime/src/main/scala/CommonRuntime.scala
@@ -16,10 +16,10 @@
 
 package examples.todolist.runtime
 
-import monix.execution.Scheduler
+import scala.concurrent.ExecutionContext
 
 trait CommonRuntime {
 
-  implicit val S: Scheduler = Scheduler.Implicits.global
+  implicit val EC: ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global
 
 }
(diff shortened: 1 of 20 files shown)


modules/examples/todolist/runtime/src/main/scala/CommonRuntime.scala
```diff
@@ -16,10 +16,10 @@
 
 package examples.todolist.runtime
 
-import monix.execution.Scheduler
+import scala.concurrent.ExecutionContext
 
 trait CommonRuntime {
 
-  implicit val S: Scheduler = Scheduler.Implicits.global
+  implicit val EC: ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global
```
JesusMtnez: If you are using `import scala.concurrent.ExecutionContext` in line 19, this default value could be shorter `ExecutionContext.Implicits.global`. :wink:

JesusMtnez:
:clap: LGTM

modules/internal/src/main/scala/client/fs2Calls.scala
```diff
@@ -60,14 +62,14 @@ object fs2Calls {
...
+      options: CallOptions)(implicit EC: ExecutionContext): Stream[F, Res] =
     monixCalls
       .bidiStreaming(
         Observable
           .fromReactivePublisher(input.toUnicastPublisher),
         descriptor,
         channel,
         options)
-      .toReactivePublisher
+      .toReactivePublisher(Scheduler(EC))
```
fedefernandez: what does the apply method? any penalizations on invoking it on each call?

juanpedromoreno: I think it's just a _decorator_ atop of the `ExecutionContext`, which would be the one used for executing all the "runnables".

Issue #228, opened 2018-04-09
Freestyle-RPC Benchmarks


juanpedromoreno:
We need to create a new module for benchmarking. https://github.com/47deg/metrifier can be used as a reference, however, we need to include all the different combinations:

* Avro, AvroWithSchema, ProtocolBuffers, etc.
* Interceptors (server, client)
* Any kind combination.

As part of this story, it would be really handy to measure and establish minimum throughput values as threshold. Hence, in CI, we might a have a new stage in the pipeline that might fail if new features/modifications included in the commits are not reaching these thresholds.

L-Lavigne:
Metrifier conversion WIP is on branch https://github.com/47deg/metrifier/commits/UpgradeToNewHotness.

PR #389, opened 2018-09-25
Auto spin-up RPC server when running benchmark

juanpedromoreno:
Related to #228 

This PR removes the need of bootstrapping an RPC server prior to run the benchmarks. In this case, an in-memory RPC server is being spun up at the beginning of the execution, and it's being shutdown at the end.


benchmarks/src/main/scala/AvroWithSchemaBenchmark.scala
```diff
@@ -23,16 +23,23 @@ import freestyle.rpc.benchmarks.Utils._
...
 class AvroWithSchemaBenchmark extends Runtime {
 
+  implicit val handler: AvroWithSchemaHandler[IO] = new AvroWithSchemaHandler[IO]
+  val sc: ServerChannel                           = ServerChannel(PersonServiceAvroWithSchema.bindService[IO])
   val client: PersonServiceAvroWithSchema.Client[IO] =
-    PersonServiceAvroWithSchema.client[IO](channel)
+    PersonServiceAvroWithSchema.clientFromChannel[IO](sc.channel)
+
+  @TearDown
+  def shutdown(): Unit = sc.shutdown(); ()
```
fedefernandez: Maybe we could write this as:
```scala
IO(sc.shutdown()).void.unsafeRunSync()
```
Making explicit the side effect

PR #409, opened 2018-10-01
Adds support for BigDecimal tagged types

### LINKED ISSUE #382: Avro4s decimal encoding
We need to encode/decode decimals based on [Avro docs](https://avro.apache.org/docs/1.8.2/spec.html#Decimal)

## Explanation
I'm going to explain the problem through an example. Suppose we have the following definition:

```avdl
record StockInfo {
  string stockId;
  decimal(10,2) price;
  decimal(5,4) rate;
}
```

Currently, we do the following:

* Transform this `avdl` in Scala classes through `sbt-avrohugger`
```scala
case class StockInfo(stockId: String, price: BigDecimal, rate: BigDecimal)
```

This is serialized/deserialized with the following:

https://github.com/frees-io/freestyle-rpc/blob/876c5f269cf9b8508d9c5c7c79cf597a8dcd8840/modules/internal/src/main/scala/encoders/avro.scala#L68-L80

In the next release, this encoders will be optional (see [`idlGenMarshallerImports` under `Plugin Settings`](https://github.com/frees-io/freestyle-rpc/blob/876c5f269cf9b8508d9c5c7c79cf597a8dcd8840/docs/src/main/tut/idl-generation.md#plugin-settings-1)) but in any case, this is not Avro compliant ([Avro docs](https://avro.apache.org/docs/1.8.2/spec.html#Decimal))

## What we want to do?
We need to find a way for inferring the scale and precision from a Scala type. The encoders and decoders are based on types.

## Solutions
Generate wrappers for BigDecimal values in order to generate the custom encoders/decoders.

### Approach 1 - Very Ugly
To have thousand of combinations of type `BigDecimalS1P1`, `BigDecimalS1P2`, `BigDecimalS1P3`, ..., `BigDecimalS2P1`, ... And they encoders/decoders. Then change sbt-avrohugger for generating these wrappers based on the `avdl` definition.

```scala
case class StockInfo(stockId: String, price: BigDecimalS10P2, rate: BigDecimalS5P4)
```

### Approach 2 - Ugly
Generate the wrappers as part of the code generation in sbt-avrohugger and in a second iteration, generate custom encoders/decoders based on the name in frees-rpc

```scala
case class BigDecimalS10P2(bd: BigDecimal)
case class BigDecimalS5P4(bd: BigDecimal)
case class StockInfo(stockId: String, price: BigDecimalS10P2, rate: BigDecimalS5P4)
```

### Approach 3 - Less ugly but still ugly
To have something like this in a avrohugger dependant library:
```scala
sealed trait Precision extends Product with Serializable
case object Precision1 extends Precision
case object Precision2 extends Precision
case object Precision3 extends Precision
// ...

sealed trait Scale extends Product with Serializable
case object Scale1 extends Scale
case object Scale2 extends Scale
case object Scale3 extends Scale
// ...
```

Make sbt-avrohugger to generate the wrapper with these types appended to the `BigDecimal` types.

```scala
case class StockInfo(
  stockId: String, 
  price: BigDecimal with Scale10.type with Precision2.type , 
  rate: BigDecimalwith Scale5.type with Precision4.type)
```

Then, we could have a custom encoders/decoders for this `BigDecimal` with extended types. This could be implemented also with shapeless.

The sbt-avrohugger customization would be behind a flag defined as a sbt setting key.

Thoughts?

fedefernandez:
Fixes #382

### DIFF
--- a/modules/internal/src/main/scala/encoders/avro.scala
+++ b/modules/internal/src/main/scala/encoders/avro.scala
@@ -68,13 +68,18 @@ object avro extends AvroMarshallers {
 
     }
 
+  @deprecated("Use `freestyle.rpc.internal.encoders.avro.bigDecimal` instead.", "0.15.1")
+  object bigdecimal extends AvroBigDecimal
+
   @deprecated(
     "This avro4s encoder doesn't fulfill the avro specs. " +
       "Be aware that this serializes without taking into account the precision and scale, " +
       "and it could lead to serialization problems if the client uses a different serializer.",
     "0.15.1"
   )
-  object bigDecimal {
+  object bigDecimal extends AvroBigDecimal
+
+  trait AvroBigDecimal {
 
     implicit object bigDecimalToSchema extends ToSchema[BigDecimal] {
       override val schema: Schema = Schema.create(Schema.Type.BYTES)
--- a/modules/internal/src/main/scala/encoders/avro.scala
+++ b/modules/internal/src/main/scala/encoders/avro.scala
@@ -107,7 +107,7 @@ object avro extends AvroMarshallers {
     }
   }
 
-  object bigdecimalTagged {
+  object bigDecimalTagged {
 
     private[this] def toDecimalTag[SP](bigDecimal: BigDecimal): BigDecimal @@ SP =
       shapeless.tag[SP][BigDecimal](bigDecimal)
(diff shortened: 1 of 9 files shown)


fedefernandez:
@juanpedromoreno 
I'm guessing since the older decimal serializer doesn't fulfill the avro specs, we can make the tagged type the default behavior for the idl gen. Let me know if that makes sense.

modules/internal/src/main/scala/encoders/avro.scala
```diff
@@ -63,7 +68,13 @@ object avro extends AvroMarshallers {
...
     }
 
-  object bigdecimal {
+  @deprecated(
+    "This avro4s encoder doesn't fulfill the avro specs. " +
+      "Be aware that this serializes without taking into account the precision and scale, " +
+      "and it could lead to serialization problems if the client uses a different serializer.",
+    "0.15.1"
+  )
+  object bigDecimal {
```
pepegar: you're breaking bincompat here by making the `D` in `Decimal` uppercase. I think it's a typo :)

PR #412, opened 2018-10-02
Migration guide for decimals

fedefernandez:
Adds the needed modules for ease the migration to the new decimal serializers.

Added a new doc page, not sure if I need to list it in any place.

I've also finished the repository with the example:
* https://github.com/frees-io/freestyle-rpc-protocol-decimal-update

### DIFF
--- a/modules/idlgen/core/src/main/scala/avro/AvroSrcGenerator.scala
+++ b/modules/idlgen/core/src/main/scala/avro/AvroSrcGenerator.scala
@@ -42,8 +42,8 @@ case class AvroSrcGenerator(
   }
   private val avroScalaCustomTypes = Standard.defaultTypes.copy(decimal = avroBigDecimal)
   private val mainGenerator        = Generator(Standard, avroScalaCustomTypes = Some(avroScalaCustomTypes))
-  private val adtGenerator = mainGenerator.copy(avroScalaCustomTypes =
-    Some(AvroScalaTypes.defaults.copy(protocol = ScalaADT))) // ScalaADT: sealed trait hierarchies
+  private val adtGenerator = mainGenerator.copy(avroScalaCustomTypes = Some(
+    mainGenerator.avroScalaTypes.copy(protocol = ScalaADT))) // ScalaADT: sealed trait hierarchies
 
   val idlType: String = avro.IdlType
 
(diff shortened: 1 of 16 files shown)


modules/idlgen/core/src/main/scala/avro/AvroSrcGenerator.scala
```diff
@@ -42,8 +42,8 @@ case class AvroSrcGenerator(
   }
   private val avroScalaCustomTypes = Standard.defaultTypes.copy(decimal = avroBigDecimal)
   private val mainGenerator        = Generator(Standard, avroScalaCustomTypes = Some(avroScalaCustomTypes))
-  private val adtGenerator = mainGenerator.copy(avroScalaCustomTypes =
-    Some(AvroScalaTypes.defaults.copy(protocol = ScalaADT))) // ScalaADT: sealed trait hierarchies
+  private val adtGenerator = mainGenerator.copy(avroScalaCustomTypes = Some(
+    mainGenerator.avroScalaTypes.copy(protocol = ScalaADT))) // ScalaADT: sealed trait hierarchies
```
fedefernandez: Maybe I should create a small PR against master to fix this.

project/ProjectPlugin.scala
```diff
@@ -223,6 +223,16 @@ object ProjectPlugin extends AutoPlugin {
...
 
+    lazy val legacyAvroDecimalProtocolSettings: Seq[Def.Setting[_]] = Seq(
+      publishMavenStyle := true,
+      crossPaths := false,
+      libraryDependencies := Nil
+    )
+
+    lazy val legacyAvroDecimalModelSettings: Seq[Def.Setting[_]] = Nil
+
+    lazy val legacyAvroDecimalEncodersSettings: Seq[Def.Setting[_]] = Nil
```
fedefernandez: We have the Scala model and the avdl type defined with the same name and the same package. In this way, the encoders can depend on the scala model for creating the serializers and the clients using avdl need to write something like this:
https://github.com/frees-io/freestyle-rpc-protocol-decimal-update/blob/master/server-v2/build.sbt#L25

Do someone knows how to exclude a "`dependsOn`" module dependency from the dependencies of that module in the pom?

fedefernandez:
@juanpedromoreno I've added the disclaimer in the migration guide and removed the `bigDecimal` object to avoid conflicts. Please, review it again when you have a chance. Thanks

PR #465, opened 2018-10-25
Dont block for server termination

tbrown1979:
#### Problem I'm trying to solve:
A lot of services I have worked with expose Grpc endpoints and http health endpoints. This results in code that looks similar to this in Main

```scala
import scala.concurrent.ExecutionContext.Implicits.global
...

for {
  _ <- ....
  exitCode <-  Stream (
    streamGrpcServer(...),
    streamHealthAndMetrics(...)
  ).join(2)
```

With the current implementation where we do:
```scala
S.start() *> shutdownEventually *> S.awaitTermination()
```
This will block 1 thread from the `ExecutionContext` used to execute the `.join(2)`. The `awaitTermination()` does not return until the `GrpcServer` is `shutdown()` which doesn't happen until the shutdown hook. I would like to prevent this project from blocking a thread while the server is running.

#### Solution:
If we move the `.awaitTermination()` to the `shutdownHook` then we're only waiting once the shutdown has already begun. This prevents us from having to block a thread to wait for shutdown. However, we still need the `server` function to not return until the server is shutdown. To satisfy this I'm using a callback to only complete the `IO.async` once the shutdown hook has finished. The `IO.async` here is sort of acting as a `Promise`/`Deferred` which won't block a thread from the `.join`'s `ExecutionContext`.


tbrown1979:
If people like this idea then I can work on fixing the tests!

juanpedromoreno:
@tbrown1979 Thanks! It looks good to me 👌

tbrown1979:
@juanpedromoreno I ended up removing a test for the `GrpcServer.server` function. The one that existed currently wasn't actually testing much of anything, from what I could tell - at least not testing the shutdown hook that needs to run. I tested this pretty thoroughly with some of the examples and things seemed to function effectively the same, besides no longer blocking a thread.

tbrown1979:
Happy to help 😄

Issue #461, opened 2018-10-19
Update dependency io.circe:circe-core


developer47deg:
Actual version: 0.9.3
 Patch: -
 Minor: 0.10.0
 Major: -

mbeatrizmagalhaes:
Hi,
I'd like to contribute by updating the dependencies of the project, per the issues. Is it ok if I open a PR with all of them updated, or would you rather one per library?

raulraja:
@mbeatrizmagalhaes thanks for the help!. Feel free to submit them all in a single PR if the changeset is small enough or separate if when updating versions you find that a lot of code also needs updating and it becomes harder to review. Thanks!

Issue #460, opened 2018-10-19
Update dependency io.netty:netty-tcnative-boringssl-static


developer47deg:
Actual version: 2.0.12.Final
 Patch: 2.0.18.Final
 Minor: -
 Major: -

Issue #456, opened 2018-10-19
Update dependency io.netty:netty-tcnative-boringssl-static:test


developer47deg:
Actual version: 2.0.12.Final
 Patch: 2.0.18.Final
 Minor: -
 Major: -

Issue #452, opened 2018-10-19
Update dependency com.sksamuel.avro4s:avro4s-core


developer47deg:
Actual version: 1.8.3
 Patch: -
 Minor: 1.9.0
 Major: 2.0.2

PR #466, opened 2018-10-28
Update dependencies: `io.circe`, `io.grpc`, netty and avro4s

### LINKED ISSUE #437: Update dependency io.circe:circe-core
Actual version: 0.9.3
 Patch: -
 Minor: 0.10.0
 Major: -

mbeatrizmagalhaes:
This PR has updates on the dependencies that didn't require code changes. From what I've understood the other ones require some more code changes that will be best on separate PRs to be more contained. 

I've updated `avro4s` to the minor version referenced on the issue (#452), but to the major also requires some bigger code changes, I believe it is also best to separate it.

This fixes #437, #438, #439, #447, #450, #453, #454 , #455, #456, #457, #458, #459, #460 and #461

### DIFF
--- a/project/ProjectPlugin.scala
+++ b/project/ProjectPlugin.scala
@@ -40,7 +40,7 @@ object ProjectPlugin extends AutoPlugin {
       val logback: String            = "1.2.3"
       val monix: String              = "3.0.0-RC1"
       val monocle: String            = "1.5.1-cats"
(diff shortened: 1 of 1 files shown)


fedefernandez:
Hi @mbeatrizmagalhaes, thanks for the PR. We can't upgrade to `avro4s` 1.9.0, there is some strange behavior with that version, you can check the details in the issue #288.

PR #471, opened 2018-11-07
Adds some Prometheus Code Improvements

juanpedromoreno:
This PR ensures that the status code value is what it's reported. Also, it adds a minor code improvement.


juanpedromoreno:
Just realized that this would break the Prometheus code in the user code, including Prometheus functions, and potentially Grafana dashboards, for example, thus I'm, closing this.

PR #477, opened 2018-11-15
Uses fs2-grpc for the fs2 integration

fedefernandez:
`fs2-grpc` doesn't have support for compression at the server level.  A PR was submitted for addressing that: https://github.com/fiadliel/fs2-grpc/pull/42.

The fs2 support is still experimental so, IMO we're fine.


fedefernandez:
Let's see. I think they aren't solved completely but at least didn't fail in the last 6/8 builds

Issue #335, opened 2018-07-02
Add a new docs section to explain models evolution compatibility


rafaparadela:
It would be helpful to explain how models can evolve, and how the serializers preserve back/forward compatibility.

As part of this new section, we can add [these conclusions for the Avro](https://github.com/frees-io/freestyle-rpc/issues/331#issuecomment-401789947) cases, but we probably should do the same thing for Protocol Buffer definitions.

fedefernandez:
@juanpedromoreno this is done, right?

juanpedromoreno:
Yes, it's partially done since we might want to add the proto section.

PR #481, opened 2018-12-09
[Docs] Schema Evolution

juanpedromoreno:
Solves https://github.com/higherkindness/mu/issues/335

### DIFF
--- a/docs/src/main/tut/schema-evolution/avro.md
+++ b/docs/src/main/tut/schema-evolution/avro.md
@@ -21,18 +21,18 @@ For Scala, this section specifies how such schema differences should be resolved
 - [Modifying the Request (Client side)](#modifying-the-request-client-side)
   - [A: Adding a new non-optional field](#a-adding-a-new-non-optional-field)
   - [B: Adding a new optional field](#b-adding-a-new-optional-field)
-  - [C: Adding new item to a coproduct](#c-adding-new-item-to-a-coproduct)
-  - [D: Removing item from a coproduct](#d-removing-item-from-a-coproduct)
-  - [E: Replacing item in a coproduct](#e-replacing-item-in-a-coproduct)
+  - [C: Adding new item to a union](#c-adding-new-item-to-a-union)
+  - [D: Removing item from a union](#d-removing-item-from-a-union)
+  - [E: Replacing item in a union](#e-replacing-item-in-a-union)
   - [F: Changing the type of an existing field](#f-changing-the-type-of-an-existing-field)
   - [G: Renaming a field](#g-renaming-a-field)
   - [H: Removing a field](#h-removing-a-field)
 - [Modifying the Response (Server side)](#modifying-the-response-server-side)
   - [I: Adding a new non-optional field](#i-adding-a-new-non-optional-field)
   - [J: Adding a new optional field](#j-adding-a-new-optional-field)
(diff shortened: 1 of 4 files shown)


docs/src/main/tut/schema-evolution/avro.md
```diff
@@ -0,0 +1,292 @@
...
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+## Cases
+
+- [Modifying the Request (Client side)](#modifying-the-request-client-side)
+  - [A: Adding a new non-optional field](#a-adding-a-new-non-optional-field)
+  - [B: Adding a new optional field](#b-adding-a-new-optional-field)
+  - [C: Adding new item to a coproduct](#c-adding-new-item-to-a-coproduct)
+  - [D: Removing item from a coproduct](#d-removing-item-from-a-coproduct)
+  - [E: Replacing item in a coproduct](#e-replacing-item-in-a-coproduct)
```
AdrianRaFo: I would say `union` instead of `coproduct` because we are talking in terms of the `avro` protocol

docs/src/main/tut/schema-evolution/avro.md
```diff
@@ -0,0 +1,292 @@
...
+
+* After:
+
+```scala
+case class NewResponse(a: String, b: Int = 123, c: Boolean)
+```
+
+#### O: Renaming a field
+
+It will require to keep the previous name and type with a default value, and then, adding a new field with the same type.
```
fedefernandez: I thought a rename shouldn't affect, probably I'm wrong.

docs/src/main/tut/schema-evolution/avro.md
```diff
@@ -0,0 +1,292 @@
...
+* Before:
+
+```scala
+case class Request(a: Int :+: String :+: CNil)
+```
+
+* After:
+
+```scala
+case class NewRequest(a: Int :+: String :+: Boolean :+: CNil)
```
AdrianRaFo: I guess this will end on a unsupported type for the server side

docs/src/main/tut/schema-evolution/avro.md
```diff
@@ -0,0 +1,292 @@
...
+
+* After:
+
+```scala
+case class NewResponse(a: String, b: Int = 123, c: Int)
+```
+
+#### P: Removing a field
+
+This evolution should never happen since we would lose backward compatibility. Nonetheless, it would work only under the special case that the old response has a default value for the field that we want to delete, where this operation would be feasible by removing the field in the new version of the server response.
```
AdrianRaFo: I understood this occurs because the client will use its default value, isn't it?

docs/src/main/tut/schema-evolution/avro.md
juanpedromoreno: Done in https://github.com/higherkindness/mu/pull/481/commits/4a79efa8d256e4cb44551a5b2f8c1aa5caa55c24

docs/src/main/tut/schema-evolution/avro.md
juanpedromoreno: You are right, I've verified it and it works like a charm for client requests and server responses.

https://github.com/higherkindness/mu/pull/481/commits/4a79efa8d256e4cb44551a5b2f8c1aa5caa55c24

PR #486, opened 2018-12-13
Downgrades sbt-org-policies and fixes codecov badge

fedefernandez:


project/ProjectPlugin.scala
```diff
@@ -62,7 +60,7 @@ object ProjectPlugin extends AutoPlugin {
         %("grpc-stub", V.grpc),
         %%("monix", V.monix),
         %%("monocle-core", V.monocle),
-        %%("fs2-reactive-streams", V.fs2),
+        "co.fs2" %% "fs2-reactive-streams" % V.fs2,
```
juanpedromoreno: Maybe I'm wrong, but can we remove this dependency now?

fedefernandez: It's used for the monix calls
https://github.com/higherkindness/mu/blob/48b8578cbafc59dc59e61e93c13fdb4709b7e540/modules/internal/src/main/scala/client/package.scala#L23

fedefernandez: Good catch

PR #487, opened 2018-12-14
Bump up gRPC version

juanpedromoreno:


juanpedromoreno:
@fedefernandez Anyway we are on 1.16.x, which will cause also an eviction with the scalapb version. I've asked for a new release in gitter though: https://gitter.im/trueaccord/ScalaPB?at=5c137b41ca317a0e25c59b4e

PR #494, opened 2018-12-18
Improve managed channel interpreter

AdrianRaFo:
This PR separates the `ManagedChannelInterpreter` class from the package object and removes the redundant parameters from its methods.

### DIFF
--- a/modules/client/src/main/scala/ManagedChannelInterpreter.scala
+++ b/modules/client/src/main/scala/ManagedChannelInterpreter.scala
@@ -23,17 +23,14 @@ import cats.syntax.functor._
 
 import io.grpc._
 
-object ManagedChannelInterpreter {
+class ManagedChannelInterpreter[F[_]](
+    initConfig: ChannelFor,
+    configList: List[ManagedChannelConfig])(implicit F: Sync[F]) {
 
-  def apply[F[_], A](
-      initConfig: ChannelFor,
-      configList: List[ManagedChannelConfig],
-      fa: ManagedChannelOps[F, A])(implicit F: Sync[F]): F[A] =
-    fa(build(initConfig, configList))
+  def apply[A](fa: ManagedChannelOps[F, A]): F[A] =
+    fa(new ManagedChannelInterpreter(initConfig, configList).build)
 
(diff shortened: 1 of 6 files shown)


modules/client/src/main/scala/ManagedChannelInterpreter.scala
```diff
@@ -0,0 +1,54 @@
...
+import cats.syntax.functor._
+
+import io.grpc._
+
+class ManagedChannelInterpreter[F[_]](
+    initConfig: ChannelFor,
+    configList: List[ManagedChannelConfig])(implicit F: Sync[F]) {
+
+  def apply[A](fa: ManagedChannelOps[F, A]): F[A] =
+    fa(new ManagedChannelInterpreter(initConfig, configList).build)
```
fedefernandez: Why not just
```scala
    def apply[A](fa: ManagedChannelOps[F, A]): F[A] =
      fa(build(initConfig, configList))
```
as before?

modules/client/src/main/scala/ManagedChannelInterpreter.scala
```diff
@@ -0,0 +1,54 @@
...
+
+    for {
+      b          <- builder
+      configured <- F.delay(configList.foldLeft(b)(configureChannel))
+      built      <- F.delay(configured.build())
+    } yield built
+  }
+
+  def unsafeBuild[T <: ManagedChannelBuilder[T]](implicit E: Effect[F]): ManagedChannel =
+    E.toIO(new ManagedChannelInterpreter(initConfig, configList).build).unsafeRunSync()
```
fedefernandez:
```scala
E.toIO(build(initConfig, configList)).unsafeRunSync()
```
???

modules/client/src/main/scala/ManagedChannelInterpreter.scala
AdrianRaFo: Because build doesn't has parameters anymore

modules/client/src/main/scala/ManagedChannelInterpreter.scala
AdrianRaFo: same as above

PR #505, opened 2018-12-28
Split streaming modules

### LINKED ISSUE #490: Splitting Mu: core + streaming modules
A good enhancement could be to split Mu into several and shorter modules in term of dependencies.

An initial proposal could be:

* **mu-core**: Depending only on `cats`/`cats-effect`.
* **mu-fs2**: depending on the core and `fs2`.
* **mu-monix**: depending on the core and `monix`.

### LINKED ISSUE #491: Unused import warning
Currently, we add all the encoder imports without taking into account if they are needed. That causes a lot of warning messages like the following:
```bash
[warn] /Users/fede/development/workspaces/47deg/mu/modules/server/src/test/scala/protocol/Utils.scala:77:6: Unused import
[warn]     @service(Protobuf) trait ProtoRPCService[F[_]] {
```

It'd be nice to find a way of hiding those messages, by only adding the import when it's necessary or by adding some flag.

fedefernandez:
Fixes #490 
Fixes #491 

This PR:
* Splits the internal and client modules in three: `core`, `monix` and `fs2` (#490)
* Reorganizes some dependencies, minimizing the evictions. 
  * The monocle utilities have been moved to idlgen-core, removing the dependency from internal
  * Downgraded cats effect to the version fs2-grpc is using
* Fixes the warning message in Protobuf services (#491)


fedefernandez:
I've tested this with the `mu-protocol-decimal-update` and works fine. Merging

PR #521, opened 2019-01-18
Passes the method info to the MetricsOps defs

fedefernandez:


modules/internal/src/main/scala/higherkindness/mu/rpc/internal/metrics/MetricsOps.scala
```diff
@@ -15,23 +15,26 @@
...
+  def recordMessageReceived(methodInfo: GrpcMethodInfo, classifier: Option[String]): F[Unit]
 
-  def recordHeadersTime(method: MethodType, elapsed: Long, classifier: Option[String]): F[Unit]
+  def recordHeadersTime(
+      methodInfo: GrpcMethodInfo,
+      elapsed: Long,
+      classifier: Option[String]): F[Unit]
 
   def recordTotalTime(
-      method: MethodType,
```
juanpedromoreno: Is the `MethodType` part of the `GrpcMethodInfo`?

fedefernandez: yes

fedefernandez: yeah, I'm almost sure this is what I did 😅

PR #522, opened 2019-01-18
Channel Interceptor Metrics

fedefernandez:
This is ready for review.

It adds the metrics interceptor for the channels.


modules/channel/src/main/scala/higherkindness/mu/rpc/channel/ManagedChannelConfig.scala
```diff
@@ -32,7 +32,11 @@ final case class OverrideAuthority(authority: String)             extends Manage
 final case class UsePlaintext()                                   extends ManagedChannelConfig
 final case class NameResolverFactory(resolverFactory: NameResolver.Factory)
     extends ManagedChannelConfig
+@deprecated(
+  "This method disables service-config-based policy selection, and may cause problems. See DefaultLoadBalancingPolicy",
+  "0.16.0")
```
juanpedromoreno: I guess this is being deprecated in the next release, potentially for `0.17.1` or `0.18.0`?

fedefernandez: You're right. Good catch

juanpedromoreno:
```suggestion
  "0.17.0")
```

Issue #514, opened 2019-01-15
Dropwizard  MetricOps implementation


fedefernandez:
Depends on #512

Add an implementation for the MetricOps based on Dropwizard.

It'll reside in a new module called `dropwizard`

See https://github.com/http4s/http4s/blob/master/dropwizard-metrics/src/main/scala/org/http4s/metrics/dropwizard/Dropwizard.scala#L67

fedefernandez:
Done in #546

PR #546, opened 2019-01-23
514 - Dropwizards implementation

jdesiloniz:
Solves #514 

This PR adds support for Dropwizard metrics for both clients and servers.

### DIFF
--- a/modules/internal/src/main/scala/higherkindness/mu/rpc/internal/metrics/MetricsOps.scala
+++ b/modules/internal/src/main/scala/higherkindness/mu/rpc/internal/metrics/MetricsOps.scala
@@ -43,7 +43,7 @@ trait MetricsOps[F[_]] {
 
 object MetricsOps {
 
-  sealed trait GrpcStatus
+  sealed trait GrpcStatus extends Product with Serialization
   case object OK                extends GrpcStatus
   case object Cancelled         extends GrpcStatus
   case object DeadlineExceeded  extends GrpcStatus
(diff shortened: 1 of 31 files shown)


modules/internal/src/main/scala/higherkindness/mu/rpc/internal/metrics/MetricsOps.scala
```diff
@@ -40,3 +40,40 @@ trait MetricsOps[F[_]] {
       classifier: Option[String]): F[Unit]
 
 }
+
+object MetricsOps {
+
+  sealed trait GrpcStatus
```
fedefernandez:
```suggestion
  sealed trait GrpcStatus extends Product with Serialization
```

jdesiloniz:
@fedefernandez Thanks! Addressed :)

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/metrics/MetricsOps.scala
```diff
@@ -40,3 +40,40 @@ trait MetricsOps[F[_]] {
       classifier: Option[String]): F[Unit]
 
 }
+
+object MetricsOps {
+
+  sealed trait GrpcStatus extends Product with Serialization
```
fedefernandez:
```suggestion
  sealed trait GrpcStatus extends Product with Serializable
```

jdesiloniz:
@fedefernandez @juanpedromoreno addressed both comments! Thanks!

Issue #182, opened 2018-02-18
HTTP Support from Protocols


juanpedromoreno:
**`frees-rpc` is for RPC**. 

Having said that, it'd be great supporting the creation of HTTP/REST endpoints taking advantage of the RPC service implementations. This is mandatory in some cases/environments when the HTTP/REST API is the only way possible.

This is related to #73, but focuses on the scenario of HTTP/REST on both ends instead of a REST<->gRPC converter or proxy.

Subtasks:

* [x] Naming. A module named `frees-rpc-http` is implemented in #203. Ideally we'll want to settle on a naming convention for the feature such as HTTP, REST or HTTP/REST. 
* Transport Implementation
  * [x] Transport/Framework layer implementation. Our current approach is to use [http4s](https://github.com/http4s/http4s) and its DSL, as implemented [here](https://github.com/frees-io/freestyle-rpc/commit/bb9f8a4d5afd139542bf3b77a1c116a60fbdcbaf).
  * [x] FS2 streaming support. Supported by http4s, mainly for server-streaming but client/bidi streaming is also possible, and implemented [here](https://github.com/frees-io/freestyle-rpc/commit/bb9f8a4d5afd139542bf3b77a1c116a60fbdcbaf#diff-1835d06e0faa670afb3c0610ecfb02ce).
  * [x] Monix/Observable streaming support. Reversing the gRPC-based approach in the main project, we can convert FS2 to Monix internally, as implemented [here](https://github.com/frees-io/freestyle-rpc/commit/3666518a658522a1d0ebfd7bb04bf009644337f4).
  * [x] Error handling. Just as with RPC, we'll want to support error raising in monadic (`MonadError`) and streaming contexts (keeping in mind that a stream can start OK and end in error). The sample implementation [here](https://github.com/frees-io/freestyle-rpc/commit/9711c2b276bf3c16c6231295b2369673901fa6e0) accepts gRPC's thrown `Status[Runtime]Exception` for triggering specific return codes, but we'll probably want to provide a more generic type of error for this.
  * [ ] HTTP/2 support: Http4s supports this on the server side (requires TLS) but the state of client-side support is less clear. This is tracked in #205.
* Endpoint Derivation
[...]

PR #552, opened 2019-01-29
Sample http4s REST client/server with client macro derivation

L-Lavigne:
Adds sample http4s REST client/server implementations, which we will eventually derive from the same `@service`-annotated classes as we do for RPC. The roadmap for this is described in #182.

Also implements HTTP client-side derivation using `@http` annotations, based on #368. Those are tested with unary clients, but support for streaming exists in our http4s samples and can be auto-derived as well. The server-side remains to be done.

Note that this client derivation is implemented inside the `serviceImpl` class that processes the `@service` annotation, due to difficulties in separating the macro processing between this class and another one for the `http` module. It still needs to be moved there.

**Warning**: do not merge in current state as this does not pass all unit tests, because auto-derived HTTP servers are not yet implemented.


rafaparadela:
This PR is ready for review. Below a brief summary about how it works:

## HTTP Protocol

Given this example protocol below:

```scala
import monix.reactive.Observable
@service(Avro) trait Greeter[F[_]] {

  @http def sayHellos(requests: Observable[HelloRequest]): F[HelloResponse]

  @http def sayHelloAll(request: HelloRequest): Observable[HelloResponse]

  @http def sayHellosAll(requests: Observable[HelloRequest]): Observable[HelloResponse]

}
```

With the annotation `@http` we would be letting the macro know that an HTTP Server should be built, which must be able to serve responses to call from the paths: `sayHellos`, `sayHelloAll`, and `sayHellosAll`.


### Run the HTTP Server

Assuming that implicits _ExecutionContext_ and _ContextShift_ are already in the scope:

```scala
implicit val ec                   = monix.execution.Scheduler.Implicits.global
implicit val cs: ContextShift[IO] = IO.contextShift(ec)
```

We need to provide an implicit instance of the interpreter of the service, that is, the `Greeter[F]` with the actual implementation of every operation.

```scala
implicit val greeterHndlerIO = new GreeterHandler[IO]
```

Then we need a server builder by:

```scala
val server = HttpServer.bind(80, "localhost", Greeter.route[IO])
```

And finally, we can run the server:

```scala
server.resource.use(_ => IO.never).start.unsafeRunSync()
```

### Run the HTTP Client

Firstly, we should set the Uri to reach the server:

```scala
val serviceUri: Uri = Uri.unsafeFromString(s"http://localhost:80")
```

Then, we can instantiate the derived client:

```scala
val client = Greeter.httpClient[IO](serviceUri)
```

And now we can perform rest calls. For instance, by:

```scala
val requests = Observable(HelloRequest("hey"), HelloRequest("there"))
val response = BlazeClientBuilder[IO](ec).resource.use(client.sayHellos(requests)(_))
```

where _response_ is `IO[HelloResponse]` as expected.

fedefernandez:
@rafaparadela this looks pretty cool. I didn't take a look at the code yet, but I have some questions/doubts:
* Is it possible now or planned for the future the ability to modify the paths? Use the method name seems a bit strict for me.
* Do we need a monix `Scheduler` or an `ExecutionContext`?
* What's the type returned by `HttpServer.bind`

rafaparadela:
@fedefernandez answering to your questions:

# Path

Initially, the `@http def sayHello` annotation was designed to accept 2 arguments `@http(POST, "sayHello") def sayHello`. The first argument was the method, that can be inferred by the type of request, and the second was the prefix to set the final path. I decided to remove those arguments and set the convention to build endpoints following this criteria: `root/serviceName/operation`. I'm assuming that these http services will be consumed by the http clients built with Mu following the same criteria, so the path, in my opinion, is not that important since are used internally. But of course I can restore the prefix argument. The example I added maybe was not very descriptive because does not show how it works with several services. So let me explain it better:

### Protocol

```scala
@message final case class HelloRequest(hello: String)
@message final case class HelloResponse(hello: String)

@service(Avro) trait UnaryGreeter[F[_]] {
  @http def getHello(request: Empty.type): F[HelloResponse]
  @http def sayHello(request: HelloRequest): F[HelloResponse]
}

import fs2.Stream
@service(Avro) trait Fs2Greeter[F[_]] {
  @http def sayHellos(requests: Stream[F, HelloRequest]): F[HelloResponse]
  @http def sayHelloAll(request: HelloRequest): Stream[F, HelloResponse]
  @http def sayHellosAll(requests: Stream[F, HelloRequest]): Stream[F, HelloResponse]
}

import monix.reactive.Observable
@service(Avro) trait MonixGreeter[F[_]] {
  @http def sayHellos(requests: Observable[HelloRequest]): F[HelloResponse]
  @http def sayHelloAll(request: HelloRequest): Observable[HelloResponse]
  @http def sayHellosAll(requests: Observable[HelloRequest]): Observable[HelloResponse]
}
```

### Server

```scala
implicit val unaryHandlerIO = new UnaryGreeterHandler[IO]
implicit val fs2HandlerIO = new Fs2GreeterHandler[IO]
implicit val monixHandlerIO = new MonixGreeterHandler[IO]

val unaryRoute: RouteMap[IO] = UnaryGreeter.route[IO]
val fs2Route: RouteMap[IO]   = Fs2Greeter.route[IO]
val monixRoute: RouteMap[IO] = MonixGreeter.route[IO]

val server: BlazeServerBuilder[IO] = HttpServer.bind(port, host, unaryRoute, fs2Route, monixRoute)
server.resource.use(_ => IO.never).start.unsafeRunSync()
```

### Client

```scala
val unaryClient = UnaryGreeter.httpClient[IO](serviceUri)
val fs2Client   = Fs2Greeter.httpClient[IO](serviceUri)
val monixClient = MonixGreeter.httpClient[IO](serviceUri)


BlazeClientBuilder[IO](ec).resource.use(unaryClient.getHello(_))
//IO[HelloResponse("hey")]

BlazeClientBuilder[IO](ec).resource.use(fs2Client.sayHellos(Stream(HelloRequest("hey"), HelloRequest("there")))(_))
//IO[HelloResponse("hey, there")]

BlazeClientBuilder[IO](ec).resource.use(monixClient.sayHellos(Observable(HelloRequest("hey"), HelloRequest("there")))(_))
//IO[HelloResponse("hey, there")]
```

So, we are binding 3 services in one single http server, but the clients need to instantiate only those http clients that are going to consume. `unaryClient`, `fs2Client` and `monixClient` bring implicitly their own paths to reach the services.

# Scheduler vs. ExecutionContext

Being technically strict, the `Scheduler` is only needed if any of the operations receives or returns a `Monix.Observable`. If no streaming (or fs2.Stream) is expected, the derived server can work with a regular execution context.

# HttpServer.bind

This is the implementation: 

```scala
def bind[F[_]: ConcurrentEffect](port: Int, host: String, routes: RouteMap[F]*): BlazeServerBuilder[F] =
    BlazeServerBuilder[F].bindHttp(port, host)
      .withHttpApp(Router(routes.map(r => (s"/${r.prefix}", r.route)): _*).orNotFound)
```
It's a ha-doc constructor for `BlazeServerBuilder[F]` which expects to receive the routes and the prefixes as `RouteMaps`:

```scala
case class RouteMap[F[_]](prefix: String, route: HttpRoutes[F])
```

which is what you have when invoke the derived routes:

```scala
val unaryRoute: RouteMap[IO] = UnaryGreeter.route[IO]
val fs2Route: RouteMap[IO]   = Fs2Greeter.route[IO]
val monixRoute: RouteMap[IO] = MonixGreeter.route[IO]
```

modules/http/src/test/scala/higherkindness/mu/rpc/http/Utils.scala
```diff
@@ -0,0 +1,125 @@
...
+import io.grpc.{Status => _, _}
+import jawnfs2._
+import monix.execution._
+import monix.reactive.Observable
+import org.http4s._
+import org.http4s.dsl.Http4sDsl
+import scala.concurrent.ExecutionContext
+import scala.util.control.NoStackTrace
+
+object Utils {
```
L-Lavigne: We need to move this class to the src/main root instead of src/test, since it will be imported in the expanded HTTP macros in production code.

We may also want to turn it into a package object.

modules/http/src/main/scala/higherkindness/mu/http/protocol.scala
```diff
@@ -0,0 +1,63 @@
...
+
+package higherkindness.mu.http
+
+import cats.effect.ConcurrentEffect
+import org.http4s.HttpRoutes
+import org.http4s.server.Router
+import org.http4s.server.blaze.BlazeServerBuilder
+import org.http4s.dsl.io._
+
+sealed trait HttpMethod extends Product with Serializable
```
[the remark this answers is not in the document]
rafaparadela: You're right.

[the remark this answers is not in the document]
rafaparadela: Done at https://github.com/higherkindness/mu/pull/552/commits/3133ec57f2375e03ae0bfa0941ca9febcfb6edab

project/ProjectPlugin.scala
```diff
@@ -34,6 +34,7 @@ object ProjectPlugin extends AutoPlugin {
       val fs2Grpc: String            = "0.4.0-M3"
       val grpc: String               = "1.18.0"
       val jodaTime: String           = "2.10.1"
+      val http4s                     = "0.19.0"
```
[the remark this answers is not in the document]
rafaparadela: I'm on it. Please note that this upgrading might affect deeply to the derivation of the server.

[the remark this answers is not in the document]
rafaparadela: Done at https://github.com/higherkindness/mu/pull/552/commits/34c05046715f0787bcf7d93554cf29ca0259b594

modules/http/src/test/scala/higherkindness/mu/rpc/http/Utils.scala
rafaparadela: Done at https://github.com/higherkindness/mu/pull/552/commits/34c05046715f0787bcf7d93554cf29ca0259b594

project/ProjectPlugin.scala
```diff
@@ -127,6 +128,21 @@ object ProjectPlugin extends AutoPlugin {
...
+        %%("http4s-dsl", V.http4s),
+        %%("http4s-blaze-server", V.http4s),
+        %%("http4s-circe", V.http4s),
+        %%("circe-generic"),
+        "co.fs2" %% "fs2-reactive-streams" % V.reactiveStreams,
+        %%("monix", V.monix),
+        %%("http4s-blaze-client", V.http4s) % Test,
+        %%("scalacheck") % Test,
+        %%("scalamockScalatest") % Test,
+        "ch.qos.logback" % "logback-classic" % "1.2.3" % Test
```
juanpedromoreno:
```suggestion
        "ch.qos.logback" % "logback-classic" % V.logback % Test
```

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
```diff
@@ -0,0 +1,124 @@
...
+import io.grpc.{Status => _, _}
+import jawnfs2._
+import monix.execution._
+import monix.reactive.Observable
+import org.http4s._
+import org.http4s.dsl.Http4sDsl
+import scala.concurrent.ExecutionContext
+import scala.util.control.NoStackTrace
+
+object Utils {
```
juanpedromoreno: I'd suggest renaming this as `implicits`, so it can be used as:

`import higherkindness.mu.http.implicits._`

Thoughts?

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
```diff
@@ -0,0 +1,124 @@
...
+            .getConstructor(classOf[String])
+            .newInstance(msg)
+            .asInstanceOf[Throwable]
+      }
+
+    def asStream[A](
+        implicit decoder: Decoder[A],
+        F: ApplicativeError[F, Throwable],
+        R: RaiseThrowable[F]): Stream[F, A] =
+      if (response.status.code != 200) Stream.raiseError(ResponseError(response.status))
```
juanpedromoreno: In order to prevent the usage of hardcoded numbers, could we use the status codes from the http4s?

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  implicit class Fs2StreamOps[F[_], A](stream: Stream[F, A]) {
+
+    implicit private val throwableEncoder: Encoder[Throwable] = new Encoder[Throwable] {
+      def apply(ex: Throwable): Json = (ex.getClass.getName, ex.getMessage).asJson
+    }
+
+    def asJsonEither(implicit encoder: Encoder[A]): Stream[F, Json] = stream.attempt.map(_.asJson)
+
+    def toObservable(implicit F: ConcurrentEffect[F], ec: ExecutionContext): Observable[A] =
```
juanpedromoreno: I wouldn't offer this combinator, in the past we used it and they turned out unsafe.

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
```diff
@@ -0,0 +1,124 @@
...
+      Observable.fromReactivePublisher(stream.toUnicastPublisher)
+  }
+
+  implicit class MonixStreamOps[A](val stream: Observable[A]) extends AnyVal {
+
+    def toFs2Stream[F[_]](implicit F: ConcurrentEffect[F], sc: Scheduler): Stream[F, A] =
```
juanpedromoreno: Same here, it's quite unsafe

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
```diff
@@ -0,0 +1,124 @@
...
+    def adaptErrors: F[Response[F]] = response.handleErrorWith {
+      case se: StatusException         => errorFromStatus(se.getStatus, se.getMessage)
+      case sre: StatusRuntimeException => errorFromStatus(sre.getStatus, sre.getMessage)
+      case other: Throwable            => InternalServerError(other.getMessage)
+    }
+
+    private def errorFromStatus(status: io.grpc.Status, message: String): F[Response[F]] =
+      status.getCode match {
+        case INVALID_ARGUMENT  => BadRequest(message)
+        case UNAUTHENTICATED   => BadRequest(message)
```
juanpedromoreno: Forbidden?

modules/http/src/test/scala/higherkindness/mu/rpc/http/GreeterRestTests.scala
```diff
@@ -0,0 +1,268 @@
...
+
+  implicit val ec                   = monix.execution.Scheduler.Implicits.global
+  implicit val cs: ContextShift[IO] = IO.contextShift(ec)
+  implicit val timer: Timer[IO]     = IO.timer(ec)
+
+  implicit val unaryHandlerIO = new UnaryGreeterHandler[IO]
+  implicit val fs2HandlerIO   = new Fs2GreeterHandler[IO]
+  implicit val monixHandlerIO = new MonixGreeterHandler[IO]
+
+  //TODO: add Logger middleware
```
juanpedromoreno: Do we need to do this _TODO_?

modules/prometheus/client/src/test/scala/higherkindness/mu/rpc/prometheus/client/BaseMonitorClientInterceptorTests.scala
```diff
@@ -327,12 +325,13 @@ abstract class BaseMonitorClientInterceptorTests extends RpcBaseTestSuite {
...
-        assertion <- check
-        _         <- serverStop[ConcurrentMonad]
-      } yield assertion).unsafeRunSync()
+        _ <- serverStart[ConcurrentMonad]
+        _ <- unary[ConcurrentMonad]
+        _ <- clientStreaming[ConcurrentMonad]
+        _ <- serverStop[ConcurrentMonad]
+      } yield ()).unsafeRunSync()
+
+      check.unsafeRunSync()
```
juanpedromoreno: These changes seem weird to me 🤔 , are we doing the check at the end? Is that correct?

project/ProjectPlugin.scala
```diff
@@ -127,6 +128,21 @@ object ProjectPlugin extends AutoPlugin {
       )
     )
 
+    lazy val httpSettings: Seq[Def.Setting[_]] = Seq(
+      libraryDependencies ++= Seq(
+        %%("http4s-dsl", V.http4s),
+        %%("http4s-blaze-server", V.http4s),
+        %%("http4s-circe", V.http4s),
+        %%("circe-generic"),
+        "co.fs2" %% "fs2-reactive-streams" % V.reactiveStreams,
```
juanpedromoreno: If we are not offering the combinators that I've mentioned above, maybe we could get rid off this dependency.

build.sbt
```diff
@@ -175,12 +182,17 @@ lazy val `dropwizard-client` = project
...
-  .settings(moduleName := "mu-rpc-dropwizard")
-  .settings(dropwizardMetricsSettings)
+///////////////////
+//// HTTP/REST ////
+///////////////////
+
+lazy val `http` = project
+  .in(file("modules/http"))
+  .dependsOn(common % "compile->compile;test->test")
+  .dependsOn(channel % "compile->compile;test->test")
```
juanpedromoreno: I think that you don't need the channel module.

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
```diff
@@ -0,0 +1,124 @@
...
+import monix.execution._
+import monix.reactive.Observable
+import org.http4s._
+import org.http4s.dsl.Http4sDsl
+import scala.concurrent.ExecutionContext
+import scala.util.control.NoStackTrace
+
+object Utils {
+
+  implicit class MessageOps[F[_]](val message: Message[F]) extends AnyVal {
```
pepegar: I'd make the value private.  Otherwise you're leaking a `message` method on all `Message[F]`:

```
val x: Message[F] = ???
x.message.message.message.message
```

```suggestion
  implicit class MessageOps[F[_]](private val message: Message[F]) extends AnyVal {
```

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
```diff
@@ -0,0 +1,124 @@
...
+
+    def jsonBodyAsStream[A](
+        implicit decoder: Decoder[A],
+        F: ApplicativeError[F, Throwable]): Stream[F, A] =
+      message.body.chunks.parseJsonStream.map(_.as[A]).rethrow
+  }
+
+  implicit class RequestOps[F[_]](val request: Request[F]) {
```
pepegar: Same here
```suggestion
  implicit class RequestOps[F[_]](private val request: Request[F]) {
```

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
```diff
@@ -0,0 +1,124 @@
...
+    def asStream[A](implicit decoder: Decoder[A], F: ApplicativeError[F, Throwable]): Stream[F, A] =
+      request
+        .jsonBodyAsStream[A]
+        .adaptError { // mimic behavior of MessageOps.as[T] in handling of parsing errors
+          case ex: ParseException =>
+            MalformedMessageBodyFailure(ex.getMessage, Some(ex)) // will return 400 instead of 500
+        }
+  }
+
+  implicit class ResponseOps[F[_]](val response: Response[F]) {
```
pepegar:
```suggestion
  implicit class ResponseOps[F[_]](private val response: Response[F]) {
```

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
```diff
@@ -0,0 +1,124 @@
...
+
+    def asStream[A](
+        implicit decoder: Decoder[A],
+        F: ApplicativeError[F, Throwable],
+        R: RaiseThrowable[F]): Stream[F, A] =
+      if (response.status.code != 200) Stream.raiseError(ResponseError(response.status))
+      else response.jsonBodyAsStream[Either[Throwable, A]].rethrow
+  }
+
+  implicit class Fs2StreamOps[F[_], A](stream: Stream[F, A]) {
```
pepegar:
```suggestion
  implicit class Fs2StreamOps[F[_], A](private stream: Stream[F, A]) {
```

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
```diff
@@ -0,0 +1,124 @@
...
+      def apply(ex: Throwable): Json = (ex.getClass.getName, ex.getMessage).asJson
+    }
+
+    def asJsonEither(implicit encoder: Encoder[A]): Stream[F, Json] = stream.attempt.map(_.asJson)
+
+    def toObservable(implicit F: ConcurrentEffect[F], ec: ExecutionContext): Observable[A] =
+      Observable.fromReactivePublisher(stream.toUnicastPublisher)
+  }
+
+  implicit class MonixStreamOps[A](val stream: Observable[A]) extends AnyVal {
```
pepegar:
```suggestion
  implicit class MonixStreamOps[A](private val stream: Observable[A]) extends AnyVal {
```

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
```diff
@@ -0,0 +1,124 @@
...
+
+    def toFs2Stream[F[_]](implicit F: ConcurrentEffect[F], sc: Scheduler): Stream[F, A] =
+      stream.toReactivePublisher.toStream[F]()
+  }
+
+  implicit class FResponseOps[F[_]: Sync](response: F[Response[F]]) extends Http4sDsl[F] {
```
pepegar:
```suggestion
  implicit class FResponseOps[F[_]: Sync](private response: F[Response[F]]) extends Http4sDsl[F] {
```

modules/prometheus/client/src/test/scala/higherkindness/mu/rpc/prometheus/client/BaseMonitorClientInterceptorTests.scala
rafaparadela: This is a fix to prevent random failures. Basically this is what I explained here: https://github.com/higherkindness/mu/pull/564

I'm not sure if after merging master this is still necessary. I'll check it out.

project/ProjectPlugin.scala
rafaparadela: done at https://github.com/higherkindness/mu/pull/552/commits/c922e0f1da911b7a61752907dc882e8522eae495

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
rafaparadela: done at https://github.com/higherkindness/mu/pull/552/commits/c922e0f1da911b7a61752907dc882e8522eae495

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
rafaparadela: Removed `toObservable` at https://github.com/higherkindness/mu/pull/552/commits/c922e0f1da911b7a61752907dc882e8522eae495

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
rafaparadela: Removed at https://github.com/higherkindness/mu/pull/552/commits/c922e0f1da911b7a61752907dc882e8522eae495

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
rafaparadela: done at https://github.com/higherkindness/mu/pull/552/commits/c922e0f1da911b7a61752907dc882e8522eae495

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
rafaparadela: replaced at https://github.com/higherkindness/mu/pull/552/commits/c922e0f1da911b7a61752907dc882e8522eae495

modules/http/src/test/scala/higherkindness/mu/rpc/http/GreeterRestTests.scala
rafaparadela: removed at https://github.com/higherkindness/mu/pull/552/commits/c922e0f1da911b7a61752907dc882e8522eae495 since we don't need it for the purpose of this test suite

project/ProjectPlugin.scala
rafaparadela: Even when we are not exposing the combinators, we still keep needing this dependency to perform some transformations.

modules/prometheus/client/src/test/scala/higherkindness/mu/rpc/prometheus/client/BaseMonitorClientInterceptorTests.scala
rafaparadela: It's still failing:

```bash
[info] - should work when combining multiple calls *** FAILED ***
[info]   1 was not equal to 
[info]   2 (BaseMonitorClientInterceptorTests.scala:321)
```

It seems we still need the fix

build.sbt
rafaparadela: Removed at https://github.com/higherkindness/mu/commit/c922e0f1da911b7a61752907dc882e8522eae495

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
rafaparadela: Done at https://github.com/higherkindness/mu/commit/c922e0f1da911b7a61752907dc882e8522eae495

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
rafaparadela: Done at https://github.com/higherkindness/mu/commit/c922e0f1da911b7a61752907dc882e8522eae495

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
rafaparadela: Done at https://github.com/higherkindness/mu/commit/c922e0f1da911b7a61752907dc882e8522eae495

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
rafaparadela: Done at https://github.com/higherkindness/mu/commit/c922e0f1da911b7a61752907dc882e8522eae495

modules/http/src/main/scala/higherkindness/mu/http/Utils.scala
rafaparadela: Done at https://github.com/higherkindness/mu/commit/c922e0f1da911b7a61752907dc882e8522eae495

rafaparadela:
Addressed all the comments by @juanpedromoreno and @pepegar

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
```diff
@@ -441,7 +441,7 @@ object serviceImpl {
 
         val executionClient: Tree = response match {
           case MonixObservableTpe(_, _) =>
-            q"client.stream(request).flatMap(_.asStream[${response.safeInner}]).toObservable"
+            q"Observable.fromReactivePublisher(client.stream(request).flatMap(_.asStream[${response.safeInner}]).toUnicastPublisher)"
```
juanpedromoreno: We need fqn for types: `_root_.monix.reactive.Observable`

The same would apply for the rest of the usages in this file.

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
```diff
@@ -338,47 +386,229 @@ object serviceImpl {
...
+		                  $responseEncoder
+		                  $requestTypology
+		                  $executionClient
+		                 }"""
+        }
+
+        val routeTypology: Tree = (request, response) match {
+          case (_: Fs2StreamTpe, _: UnaryTpe) =>
+            q"""val requests = msg.asStream[${operation.request.safeInner}]
+              Ok(handler.${operation.name}(requests).map(_.asJson))"""
```
juanpedromoreno: In all the macros stuff we need FQN

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
```diff
@@ -338,47 +386,229 @@ object serviceImpl {
...
+              responses <- Ok(handler.${operation.name}(request).asJsonEither)
+            } yield responses"""
+
+          case (_: Fs2StreamTpe, _: Fs2StreamTpe) =>
+            q"""val requests = msg.asStream[${operation.request.safeInner}]
+             Ok(handler.${operation.name}(requests).asJsonEither)"""
+
+          case (_: MonixObservableTpe, _: UnaryTpe) =>
+            q"""val requests = msg.asStream[${operation.request.safeInner}]
+              Ok(handler.${operation.name}(Observable.fromReactivePublisher(requests.toUnicastPublisher)).map(_.asJson))"""
```
juanpedromoreno: I'm not pretty sure why are converting back and forth the streaming types (Observable <-> fs2.Stream).

Could we use `higherkindness.mu.rpc.internal.client.fs2Calls`, `higherkindness.mu.rpc.internal.server.fs2Calls`, `higherkindness.mu.rpc.internal.client.monixCalls` and `higherkindness.mu.rpc.internal.server.monixCalls`?

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
```diff
@@ -338,47 +386,229 @@ object serviceImpl {
...
+        q"import _root_.org.http4s._",
+        q"import _root_.org.http4s.circe._",
+        q"import _root_.io.circe._",
+        q"import _root_.io.circe.generic.auto._",
+        q"import _root_.io.circe.syntax._"
+      )
+
+      val scheduler: List[Tree] = operations
+        .find(_.operation.isMonixObservable)
+        .map(_ => q"import _root_.monix.execution.Scheduler.Implicits.global")
```
juanpedromoreno: The execution context needs to be asked as implicit parameter. We cannot add the global here.

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
```diff
@@ -338,47 +386,229 @@ object serviceImpl {
...
+
+      val httpImports: List[Tree] = List(
+        q"import _root_.higherkindness.mu.http.implicits._",
+        q"import _root_.fs2.interop.reactivestreams._",
+        q"import _root_.cats.syntax.flatMap._",
+        q"import _root_.cats.syntax.functor._",
+        q"import _root_.org.http4s._",
+        q"import _root_.org.http4s.circe._",
+        q"import _root_.io.circe._",
+        q"import _root_.io.circe.generic.auto._",
```
fedefernandez: IMO we should ask for the encoders/decoders implicitly and if the users want auto derivation they just need to import this.

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
L-Lavigne: For context, I created this part while we still used the old conversions for the RPC calls (before https://github.com/higherkindness/mu/pull/477), except that RPC had Observable as the underlying type and FS2 conversions to it, while HTTP had the reverse. So we might be able to leverage the new RPC approach here.

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
rafaparadela: That makes sense. I'm on it right now but isn't being trivial, since the class now has a variable number of params (depending on the requests/responses).

rafaparadela: [At this commit](https://github.com/higherkindness/mu/pull/552/commits/9b0a8f3c29fe1b00b30c08db8cb695f5bbdaf252#diff-512de2cbb83d64082917cbe6d941be08L559), the implicit constraints for all the encoders and decoders have been propagated down the road, so the app who wants to run the server needs to provide implicit decoders for requests and encoders for responses, in the same way, the clients that want to perform a call need to provide implicit encoders for request and decoders for responses.

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
rafaparadela: Done at https://github.com/higherkindness/mu/pull/552/commits/9b0a8f3c29fe1b00b30c08db8cb695f5bbdaf252

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
rafaparadela: Done at https://github.com/higherkindness/mu/pull/552/commits/9b0a8f3c29fe1b00b30c08db8cb695f5bbdaf252

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
rafaparadela: We'll tackle this in the next iteration if both of you agree

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
rafaparadela: Moved at https://github.com/higherkindness/mu/pull/552/commits/fb91483df7f253ee23f22d6e6da68318b2e9e6d6

project/ProjectPlugin.scala
```diff
@@ -127,6 +128,21 @@ object ProjectPlugin extends AutoPlugin {
       )
     )
 
+    lazy val httpSettings: Seq[Def.Setting[_]] = Seq(
+      libraryDependencies ++= Seq(
+        %%("http4s-dsl", V.http4s),
+        %%("http4s-blaze-server", V.http4s),
+        %%("http4s-circe", V.http4s),
+        %%("circe-generic"),
```
juanpedromoreno:
```suggestion
        %%("circe-generic") % Test,
```

project/ProjectPlugin.scala
```diff
@@ -127,6 +128,21 @@ object ProjectPlugin extends AutoPlugin {
...
+        "co.fs2" %% "fs2-reactive-streams" % V.reactiveStreams,
+        %%("monix", V.monix),
+        %%("http4s-blaze-client", V.http4s) % Test,
+        %%("scalacheck") % Test,
+        %%("scalamockScalatest") % Test,
```
juanpedromoreno:
```suggestion
```

modules/http/src/main/scala/higherkindness/mu/http/implicits.scala
```diff
@@ -0,0 +1,128 @@
...
+
+    implicit def EitherDecoder[A, B](
+        implicit a: Decoder[A],
+        b: Decoder[B]): Decoder[Either[A, B]] = {
+      val l: Decoder[Either[A, B]] = a.map(Left.apply)
+      val r: Decoder[Either[A, B]] = b.map(Right.apply)
+      l or r
+    }
+
+    implicit private val throwableDecoder: Decoder[Throwable] =
```
fedefernandez: Please, help me to understand this. This is only used for streams and observable, right? The server generates an Either that is parsed in the client. Am I right?

modules/http/src/main/scala/higherkindness/mu/http/implicits.scala
```diff
@@ -0,0 +1,128 @@
...
+      Decoder.decodeTuple2[String, String].map {
+        case (cls, msg) =>
+          Class
+            .forName(cls)
+            .getConstructor(classOf[String])
+            .newInstance(msg)
+            .asInstanceOf[Throwable]
+      }
```
fedefernandez: Since we're losing the stacktrace here, I'd use our custom `case class`:
```scala
final case class UnexpectedError(status: Status, msg: Option[String] = None)
    extends RuntimeException(status + msg.fold("")(": " + _))
    with NoStackTrace 
//...
implicit val unexpectedErrorDecoder: Decoder[UnexpectedError] = deriveDecoder
```

modules/http/src/main/scala/higherkindness/mu/http/implicits.scala
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  implicit class Fs2StreamOps[F[_], A](private val stream: Stream[F, A]) {
+
+    implicit def EitherEncoder[A, B](implicit ea: Encoder[A], eb: Encoder[B]): Encoder[Either[A, B]] =
+      new Encoder[Either[A, B]] {
+        final def apply(a: Either[A, B]): Json = a match {
+          case Left(a)  => a.asJson
+          case Right(b) => b.asJson
+        }
```
fedefernandez:
```scala
new Encoder[Either[A, B]] {
  final def apply(a: Either[A, B]): Json = a.fold(_.asJson, _.asJson)
}
```

modules/http/src/main/scala/higherkindness/mu/http/implicits.scala
```diff
@@ -0,0 +1,128 @@
...
+      }
+
+    implicit val throwableEncoder: Encoder[Throwable] = new Encoder[Throwable] {
+      def apply(ex: Throwable): Json = (ex.getClass.getName, ex.getMessage).asJson
+    }
```
fedefernandez: Based on the above:
```scala
implicit val unexpectedErrorEncoder: Encoder[UnexpectedError] = deriveEncoder
```

modules/http/src/main/scala/higherkindness/mu/http/implicits.scala
```diff
@@ -0,0 +1,128 @@
...
+
+    def asJsonEither(implicit encoder: Encoder[A]): Stream[F, Json] = stream.attempt.map(_.asJson)
```
fedefernandez:
```scala
def asJsonEither(implicit encoder: Encoder[A]): Stream[F, Json] = 
  stream.attempt.bimap(e => UnexpectedError(e.getClass.getName, Option(e.getMessage)), _.asJson)
```

modules/http/src/main/scala/higherkindness/mu/http/implicits.scala
fedefernandez: IMO we need to avoid custom logic in the encoders/decoders

modules/http/src/main/scala/higherkindness/mu/http/implicits.scala
rafaparadela: The `implicits.scala` was enterally created by @L-Lavigne, so I'll let him answer all your questions.

modules/http/src/main/scala/higherkindness/mu/http/implicits.scala
rafaparadela: I'm sorry @fedefernandez, I don't really understand how you expect to instantiate the `UnexpectedError` without having the `Status`.

Maybe you're proposing `case class UnexpectedError(status: String, msg: Option[String] = None)`, instead of `case class UnexpectedError(status: Status, msg: Option[String] = None)`?

rafaparadela:
@fedefernandez Please, have a look at [implicits.scala](https://github.com/higherkindness/mu/blob/bf6e85319ca06d9a3b42be4fb4275bd06b7374c9/modules/http/src/main/scala/higherkindness/mu/http/implicits.scala) again and let me know if now makes more sense to you. Thanks

modules/http/src/main/scala/higherkindness/mu/http/implicits.scala
```diff
@@ -121,8 +116,18 @@ object implicits {
...
       ResponseError(errorResponse.status, Some(body).filter(_.nonEmpty)))
+
+  implicit class ThrowableOps(self: Throwable) {
+    def toUnexpected: UnexpectedError =
+      UnexpectedError(self.getClass.getName, Option(self.getMessage))
+  }
+
 }
 
+final case class UnexpectedError(className: String, msg: Option[String])
```
L-Lavigne: The name is a bit generic - most errors are unexpected. The way we're using those, they're closer to "RequestException"s, or 4XX-status `ResponseException`s. Can we combine those error types or rename this one?

modules/http/src/test/scala/higherkindness/mu/rpc/http/GreeterRestTests.scala
```diff
@@ -254,15 +250,5 @@ class GreeterRestTests
         .getOrElse(sys.error("Stuck!")) shouldBe Nil
     }
 
-    "serve ScalaCheck-generated POST requests with bidirectional Observable streaming" in {
```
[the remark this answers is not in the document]
L-Lavigne: (resolved offline)

modules/http/src/main/scala/higherkindness/mu/http/implicits.scala
L-Lavigne: (resolved offline)

L-Lavigne:
As discussed offline, I approve PR (can't set it to that status as I'm the owner), and I'll make a separate PR to adjust the naming of the exceptions.
(@rafaparadela  please let me know if you don't have merge permissions and then I'll do it)

PR #556, opened 2019-01-31
Restores the sbt plugin release and fixes ci release

fedefernandez:


version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.17.1"
+version in ThisBuild := "0.17.1-SNAPSHOT"
```
juanpedromoreno:
```suggestion
version in ThisBuild := "0.17.2-SNAPSHOT"
```

.travis.yml
```diff
@@ -58,7 +58,9 @@ jobs:
               git clean -f;
               git checkout master;
               git pull origin master;
+              sbt ";project idlgen-sbt;release";
               sbt release;
               sbt docs/publishMicrosite;
             fi
+            if [ $TRAVIS_SCALA_VERSION = "2.12.8" ]; then sbt ";project legacy-avro-decimal-compat-protocol;release"; fi
```
juanpedromoreno: Is this needed? Aren't we already in the 2.12.8 branch? https://github.com/higherkindness/mu/pull/556/files#diff-354f30a63fb0907d4ad57269548329e3R48

fedefernandez: Oh, that's true

juanpedromoreno: Probably `crossScalaVersions := Seq("2.12.8")` should be fine for that specific module

PR #566, opened 2019-03-02
RPC Server Start with Brackets

AdrianRaFo:
Use F bracket on GrpcServer


fedefernandez:
@adrianrafo the current implementation doesn't block the thread. The awaitTermination() does not return until the GrpcServer is shutdown() which doesn't happen until the shutdown hook. Have you checked if this is the case with this approach?

AdrianRaFo:
Due to brack impl bock the thread I'm going just to delete the comment because it's not needed anymore

PR #572, opened 2019-03-07
Prometheus MetricsOps implementation

jdesiloniz:
This PR adds an implementation for the MetricsOps algebra targetting Prometheus metrics.

Thanks @fedefernandez for your help with the metrics tests!


modules/metrics/prometheus/src/main/scala/higherkindness/mu/rpc/prometheus/PrometheusMetrics.scala
```diff
@@ -16,12 +16,172 @@
...
+ *
+ * {prefix}_active_calls{labels=classifier} - Counter
+ * {prefix}_messages_sent{labels=classifier,service,method} - Counter
+ * {prefix}_messages_received{labels=classifier,service,method} - Counter
+ * {prefix}_calls_header{labels=classifier} - Histogram
+ * {prefix}_calls_total{labels=classifier} - Histogram
+ * {prefix}_calls_by_method{labels=classifier,method} - Histogram
+ * {prefix}_calls_by_method{labels=classifier,status} - Histogram
+ *
+ * `methodType` can be one of the following:
```
fedefernandez:
```suggestion
 * `method` can be one of the following:
```

modules/metrics/prometheus/src/main/scala/higherkindness/mu/rpc/prometheus/PrometheusMetrics.scala
```diff
@@ -16,12 +16,172 @@
...
+ *
+ * The list of registered metrics contains:
+ *
+ * {prefix}_active_calls{labels=classifier} - Counter
+ * {prefix}_messages_sent{labels=classifier,service,method} - Counter
+ * {prefix}_messages_received{labels=classifier,service,method} - Counter
+ * {prefix}_calls_header{labels=classifier} - Histogram
+ * {prefix}_calls_total{labels=classifier} - Histogram
+ * {prefix}_calls_by_method{labels=classifier,method} - Histogram
+ * {prefix}_calls_by_method{labels=classifier,status} - Histogram
```
fedefernandez: I think we should combine these two metrics into a single one, just adding the three labels (`classifier`, `method`, and `status`)

PR #579, opened 2019-03-18
Adapts latest changes in the code generation by Skeuomorph

### LINKED ISSUE #578: Propagates imports instead of cloning dependent messages
With https://github.com/higherkindness/skeuomorph/issues/74 done. We need to adapt these changes in Mu.

In summary:

If we have `author.proto`:

```proto
package com.acme;

message Author {
    string name = 1;
    string nick = 2;
}
```

and a `book.proto`:

```proto
package com.acme;

import "author.proto";

message Book {
    int64 isbn = 1;
    string title = 2;
    repeated Author author = 3;
}
```

When we compile `book.proto`:

## What we had previously

We are generating code like this:

```scala
package com.acme

object book {
  @message final case class Author(name: String, nick: String)
  @message final case class Book(isbn: Long, title: String, author: List[Author])
}
```

## What we have now

We are generating code like this:

```scala
package com.acme

import com.acme.author.Author

object book {
  @message final case class Book(isbn: Long, title: String, author: List[Author])
}
```

rafaparadela:
This PR closes #578 

Must be in hold for now until having the Skeuomprh change released

### DIFF
--- a/project/ProjectPlugin.scala
+++ b/project/ProjectPlugin.scala
@@ -50,7 +50,7 @@ object ProjectPlugin extends AutoPlugin {
       val scalacheckToolbox: String  = "0.2.5"
       val scalamockScalatest: String = "3.6.0"
       val scalatest: String          = "3.0.6"
-      val skeuomorph: String         = "0.0.5"
+      val skeuomorph: String         = "0.0.6-SNAPSHOT"
       val slf4j: String              = "1.7.26"
       val dropwizard: String         = "4.0.5"
     }
--- a/project/ProjectPlugin.scala
+++ b/project/ProjectPlugin.scala
@@ -50,7 +50,7 @@ object ProjectPlugin extends AutoPlugin {
       val scalacheckToolbox: String  = "0.2.5"
       val scalamockScalatest: String = "3.6.0"
       val scalatest: String          = "3.0.6"
-      val skeuomorph: String         = "0.0.6"
+      val skeuomorph: String         = "0.0.7"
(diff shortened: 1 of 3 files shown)


project/ProjectPlugin.scala
```diff
@@ -50,7 +50,7 @@ object ProjectPlugin extends AutoPlugin {
       val scalacheckToolbox: String  = "0.2.5"
       val scalamockScalatest: String = "3.6.0"
       val scalatest: String          = "3.0.6"
-      val skeuomorph: String         = "0.0.5"
+      val skeuomorph: String         = "0.0.6-SNAPSHOT"
```
juanpedromoreno:
```suggestion
      val skeuomorph: String         = "0.0.6"
```

project/ProjectPlugin.scala
```diff
@@ -50,7 +50,7 @@ object ProjectPlugin extends AutoPlugin {
       val scalacheckToolbox: String  = "0.2.5"
       val scalamockScalatest: String = "3.6.0"
       val scalatest: String          = "3.0.6"
-      val skeuomorph: String         = "0.0.5"
+      val skeuomorph: String         = "0.0.6"
```
juanpedromoreno:
```suggestion
      val skeuomorph: String         = "0.0.7"
```

juanpedromoreno: You could apply this change once https://github.com/higherkindness/skeuomorph/pull/77 is merged and released

PR #581, opened 2019-03-19
Skipping binary dependencies when is not necessary

### LINKED ISSUE #575: Skipping binary dependencies for the HTTP capabilities
During the code review of the first iteration of the HTTP feature, @juanpedromoreno suggested:

- Create another `@service` macro in the `http` module to derive both client and server in this module, thus only the users who add `mu-http` will bring important binary dependencies (like http4s and circe) to their project.

- So, if this happens, the @http annotation might be no longer necessary, since adding the mentioned new macro could be the only required step to trigger the derivation.

rafaparadela:
This PR closes #575 

Now, if you have this service:

```scala
@service(Avro) trait UnaryGreeter[F[_]] {
  def getHello(request: Empty.type): F[HelloResponse]
  def sayHello(request: HelloRequest): F[HelloResponse]
}
```

Means that you don't want to have any http capability for your service, so `@service` is generating no http stuff. So no http4s dependency is necessary.

But if you add `@http` to any operation, like that:

```scala
@service(Avro) trait UnaryGreeter[F[_]] {
  @http def getHello(request: Empty.type): F[HelloResponse]
  def sayHello(request: HelloRequest): F[HelloResponse]
}
```

Given that the `@http` annotation has been moved to `http` module, you will need to add the `mu-http` (with the transitive http4s dependency).


modules/http/src/main/scala/higherkindness/mu/http/package.scala
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package higherkindness.mu
+import scala.annotation.StaticAnnotation
+
+package object http {
```
juanpedromoreno: Why do we need a package object here instead of a standard one?

modules/http/src/test/scala/higherkindness/mu/rpc/http/GreeterServices.scala
```diff
@@ -17,6 +17,7 @@
 package higherkindness.mu.rpc.http
 
 import higherkindness.mu.rpc.protocol._
+import higherkindness.mu.http._
```
juanpedromoreno: Maybe for consistency, we want to have an additional subpackage `protocol`, like in the `rpc` case.

rafaparadela: Done at https://github.com/higherkindness/mu/pull/581/commits/b8be4b74924844d2ecddcb626dd2477db816207c

modules/http/src/main/scala/higherkindness/mu/http/package.scala
rafaparadela: Done at 
https://github.com/higherkindness/mu/pull/581/commits/b8be4b74924844d2ecddcb626dd2477db816207c

modules/http/src/main/scala/higherkindness/mu/http/protocol.scala
```diff
@@ -21,6 +21,13 @@ import org.http4s.HttpRoutes
 import org.http4s.server.blaze.BlazeServerBuilder
 import org.http4s.implicits._
 import org.http4s.server.Router
+import scala.annotation.StaticAnnotation
+
+object protocol {
```
juanpedromoreno: I was thinking about an standard package rather than an object

rafaparadela: Take a look at the name of the file, we already had `protocol.scala` (because this was cloned from the rpc part).

PR #584, opened 2019-03-25
Remove the ExecutionContext implicit when not necessary

fedefernandez:
Since we usually have a `ConcurrentEffect` and a `ContextShift` for running the apps, this uses the `ContextShift` instead of the `ExecutionContext`.

In the case where we need a `monix` `Scheduler`, we ask for him explicitly. This allows to customize and cache the `Scheduler` in the program and removes the necessity of building the object internally whenever is needed.


fedefernandez:
@juanpedromoreno ready for a second look. Thanks

PR #587, opened 2019-04-01
Removes the compatibility of Monix.Observable in the HTTP layer

### LINKED ISSUE #574: Monix.Observable <> Fs2.Stream
As discussed in this [thread](https://github.com/higherkindness/mu/pull/552#discussion_r262914248), we want to avoid using `fs2-reactive-streams` to convert Streams into Observables and vice-versa, in the context of the http server and client.

Instead of this, we should use any of the implemented solutions at:

- `higherkindness.mu.rpc.internal.client.fs2Calls`
- `higherkindness.mu.rpc.internal.server.fs2Calls`
- `higherkindness.mu.rpc.internal.client.monixCalls`
- `higherkindness.mu.rpc.internal.server.monixCalls`

rafaparadela:
Just this. Close #574

### DIFF
--- a/modules/http/src/test/scala/higherkindness/mu/rpc/http/GreeterRestTests.scala
+++ b/modules/http/src/test/scala/higherkindness/mu/rpc/http/GreeterRestTests.scala
@@ -18,14 +18,12 @@ package higherkindness.mu.rpc.http
 
 import cats.effect.{IO, _}
 import fs2.Stream
-import fs2.interop.reactivestreams._
 import higherkindness.mu.http.{ResponseError, UnexpectedError}
 import higherkindness.mu.rpc.common.RpcBaseTestSuite
 import higherkindness.mu.http.implicits._
 import io.circe.Json
 import io.circe.generic.auto._
 import io.circe.syntax._
-import monix.reactive.Observable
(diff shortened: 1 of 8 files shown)


modules/http/src/test/scala/higherkindness/mu/rpc/http/GreeterRestTests.scala
```diff
@@ -48,25 +44,21 @@ class GreeterRestTests extends RpcBaseTestSuite with BeforeAndAfter {
   val Fs2ServicePrefix   = "Fs2Greeter"
   val MonixServicePrefix = "MonixGreeter"
 
-  implicit val ec                   = monix.execution.Scheduler.Implicits.global
+  implicit val ec                   = scala.concurrent.ExecutionContext.Implicits.global
```
fedefernandez: Do you know who requires an implicit `ExecutionContext`? With the `ContextShift` it shouldn't be necessary to be implicit.

rafaparadela: Merged. I’m clarifying this in my current branch. Thanks.

PR #600, opened 2019-04-15
Update docs

AdrianRaFo:
Fix some mistakes and deprecations on docs.
Protobuf `srcGen` support added to docs.
Added a PR template to check docs update.


docs/src/main/tut/idl-generation.md
```diff
@@ -201,7 +213,8 @@ import higherkindness.mu.rpc.protocol._
 
 @message case class HelloResponse(reply: String)
 
-@service trait GreeterService[F[_]] {
+@service(Avro)
```
juanpedromoreno: Maybe, to prevent this falls outdated again, would it make sense if this block is compiled with `tut`?

sander:
When will this be published to http://higherkindness.io/mu/ ?

juanpedromoreno:
👍 Done @sander !

Issue #599, opened 2019-04-15
Feature request: use idiomatic gRPC endpoint URLs


sander:
Currently, when defining:

```scala
package com.example.ns {
  @outputPackage("com.example.grpcns")
  object protocol {
    // ...
    @service(Protobuf) {
    trait MyService[F[_]] {
      def myMethod(request: MyRequest): F[MyResponse]
    }
  }
}
```

the `idlGen` command produces idiomatic gRPC code:

```proto
// ...
package com.example.grpcns;
// ...
service MyService {
  rpc MyMethod (MyRequest) returns (MyResponse);
}
```

When using `protoc` to generate client code for this, the client calls `com.example.grpcns.MyService/MyMethod` as expected:

* with namespace
* with UpperCamelCase method name

But the Mu server will only accept `MyService/myMethod` calls:

* without a namespace
* with lowerCamelCase method name

This seems to be non-idiomatic gRPC usage and breaks compatibility with IDL-generated clients.

Suggestions for the served gRPC endpoint URLs:

1. Use the @outputPackage annotation to define the namespace.
2. Capitalize the first letter of the method name.

Example project:

- https://github.com/sander/rpctest/tree/f45616a720091d88c06d0a5be00f5a2f4306f7aa doesn’t work, since the Go client tried to reach a prefixed and UpperCamelCase method
- https://github.com/sander/rpctest/tree/e018d5e5d1812b107d976aa7bbfb6617c223c761 does work, but now there is no namespace support and I needed to write UpperCamelCase in Scala which seems un-idiomatic

Chicker:
I faced the same problem during implementing grpc health checks (see https://github.com/grpc/grpc/blob/master/doc/health-checking.md) for our service deployed in Kubernetes. According to the doc above we need to specify concrete namespace `grpc.health.v1` of our implementation of the health check service. But, I don’t know how to do it. I tried to do something like this:

```
package grpc.health.v1

import higherkindness.mu.rpc.protocol._

@service(Protobuf)
trait Health[F[_]] {
  def Check(request: ReqHealthCheck): F[RespHealthCheck]
}
```

But it didn’t help me. I get the same error: `grpc-message: Method not found: grpc.health.v1.Health/Check`

@sander  Did you solve the problem of specifying namespace somehow?

sander:
@Chicker not yet. @fedefernandez said in Gitter chat that he’d try to prioritize this issue.

Haven’t tried, but maybe this works as a workaround?

```scala
[...]

PR #601, opened 2019-04-22
Allows specifying the namespace and the capitalize params

fedefernandez:
Relates to #599

This can be considered the first step where we allow to specify a namespace and the ability to capitalize the gRPC method


fedefernandez:
Thanks @rafaparadela. The namespace is for the service and the `Capitalize` option is for the method, so I prefer to have those separated in the annotation. I'll provide only one setting key at plugin level, because as you said they are related.

PR #602, opened 2019-04-23
Improves how the params are received in the annotation

fedefernandez:
## What this does?
Allows specifying named params in the service annotation

## Checklist


fedefernandez:
@juanpedromoreno with these changes, the scenario in #599 is covered since it's using the scala file for generating the proto file. See https://github.com/sander/rpctest/compare/master...fedefernandez:ff/update-namespace?expand=1

I would want to create two following PRs. One with a new setting in the idl plugin for adding those annotations params and another one for the docs. Does it make sense?

PR #615, opened 2019-05-29
Fix compile errors with the generated source code

noelmarkham:
## What this does?
  - The "Cop" regex was not picking up the new Option[A] types
  - The "fs2" prefix on Stream was causing the macro to generate the wrong code

## Checklist

### DIFF
--- a/modules/idlgen/core/src/main/scala/higherkindness/mu/rpc/idlgen/proto/ProtoSrcGenerator.scala
+++ b/modules/idlgen/core/src/main/scala/higherkindness/mu/rpc/idlgen/proto/ProtoSrcGenerator.scala
@@ -51,7 +51,7 @@ object ProtoSrcGenerator extends SrcGenerator {
       case a      => a
     }).mkString("\n")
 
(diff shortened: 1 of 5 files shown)


noelmarkham:
@fedefernandez I'm wondering if you're able to help with the Travis failure? https://travis-ci.org/higherkindness/mu/jobs/538794629#L1284

I think it's failing because of the `fs2.` prefix that was added to Skeuomorph, which is going out as part of this release, but I'm not 100% how to debug and fix this one. My original commit fixed one macro where the `fs2.` prefix wasn't on a String so a pattern match further on was matching a different case. It may be easier to roll back the original `fs2.` prefix from Skeuomorph, but I don't know the reason for adding it in the first place.

fedefernandez:
hey @noelmarkham, I guess you're referring to this one:
https://github.com/higherkindness/skeuomorph/pull/84
The idea behind that was to avoid the import, especially painful when we don't have any stream operations and still adding the import, causing the wart plugin throwing warnings or errors at compile time.

Honestly, I don't know from the top of my head why we're getting this compilation error. I'll try to find some time later to have a look, but we can revert my PR and go back with the previous code as the lesser of two evils.

noelmarkham:
Nice one, thanks - I'll update Skeuomorph and push out a new release for that

Issue #606, opened 2019-04-25
Default values sent by third party clients fails in rpc-server.


maurogonzalez:
I create a `protocol.proto` file with a numeric type in a message
```
syntax = "proto3";

package rpc.pkg;


message HelloRequest {
  float name = 1;
}

message HelloResponse {
  string message = 1;
}

service Greeter {
  rpc SayHello (HelloRequest) returns (HelloResponse);
}
```
 but server throws the following error while calling from a grpc client:

```
Apr 25, 2019 12:12:26 PM io.grpc.internal.SerializingExecutor run
SEVERE: Exception while executing runnable io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener$1MessagesAvailable@5544a4d2
java.util.NoSuchElementException: head of empty list
        at scala.collection.immutable.Nil$.head(List.scala:426)
        at scala.collection.immutable.Nil$.head(List.scala:423)
        at pbdirect.PBParserImplicits.$anonfun$requiredParser$1(PBReader.scala:173)
        at pbdirect.PBParserImplicits.$anonfun$requiredParser$1$adapted(PBReader.scala:162)
        at pbdirect.LowPriorityPBParserImplicits$$anon$2.parse(PBReader.scala:119)
        at pbdirect.LowPriorityPBParserImplicits.$anonfun$consParser$1(PBReader.scala:128)
        at pbdirect.LowPriorityPBParserImplicits.$anonfun$consParser$1$adapted(PBReader.scala:127)
        at pbdirect.LowPriorityPBParserImplicits$$anon$2.parse(PBReader.scala:119)
        at pbdirect.PBReaderImplicits.$anonfun$prodReader$1(PBReader.scala:62)
        at pbdirect.LowerPriorityPBReaderImplicits$$anon$1.read(PBReader.scala:38)
        at pbdirect.PBParserImplicits.$anonfun$requiredParser$1(PBReader.scala:169)
        at pbdirect.PBParserImplicits.$anonfun$requiredParser$1$adapted(PBReader.scala:162)
        at pbdirect.LowPriorityPBParserImplicits$$anon$2.parse(PBReader.scala:119)
        at pbdirect.package$PBParserOps$.pbTo$extension(package.scala:47)
        at higherkindness.mu.rpc.internal.encoders.pbd$$anon$1.parse(pbd.scala:40)
        at io.grpc.MethodDescriptor.parseRequest(MethodDescriptor.java:289)
[...]

PR #616, opened 2019-06-03
POC: Filling in a default value when the client does not send it to the server

noelmarkham:
I'd love some input on this please.

Looking into #606, it is possible for clients not to send values to the server if they are the default values, and Mu currently doesn't handle that.

I've taken a rough stab at how this could be resolved, expecting the deserialised type to have a `Monoid` instance, and then using the `empty` element for that should the value not be present. By wrapping the type to be deserialised in an `Option`, this ends up calling a different implicit parser which does not explode with the `head of empty list` exception, instead returning a None.

However, the biggest question I have about this is, requiring a `Monoid` then "infects" any service creation, requiring all of the types to have a `Monoid` instance. This makes sense on the one hand, but it does seem a little heavy-handed to me. I'm not entirely sure of a way around this, if there are any cleaner suggestions I'd love to hear it.

The majority of this PR is just making this all compile with the `Monoid` requirement, I started with some automatic derivation using `kittens` but in the end I was just using `???` to speed this up. The actual code of interest is in `modules/internal/src/main/scala/higherkindness/mu/rpc/internal/encoders/pbd.scala`

Creating a server with this code does fix the issue originally submitted in #606.


toddburnside:
Using a Monoid does seem a bit heavy if you are only using `empty` and not `combine`. Do you think it's worth creating some sort of Defaultable typeclass for this?

noelmarkham:
> Using a Monoid does seem a bit heavy if you are only using `empty` and not `combine`. Do you think it's worth creating some sort of Defaultable typeclass for this?

This did cross my mind. The same issues concerning this ticket would still prevail though, it would just be a different typeclass scattered throughout.

toddburnside:
The infection part is still true, of course. But, you wouldn't have to worry about having an associative operation for the data types. Even if there is one, it is irrelevant for this usage.

noelmarkham:
> The infection part is still true, of course. But, you wouldn't have to worry about having an associative operation for the data types. Even if there is one, it is irrelevant for this usage.

Yep, agreed

pepegar:
There's `Empty` in alleycats, if you don't want to create your own. It's basically `Monoid % Semigroup` :)

fedefernandez:
@noelmarkham I like it the approach my main concern is if the `Monoid` will match the proto3 specs related to default values, but I guess it'll. 

Related to infect the code, I think it's the prize to pay if we want to fail at compilation time. Since the current error is a runtime error in the `parse` def, another approach would be to try to find a default value inside the `parse`  and throw an error in case we can't find it.

diesalbla:
> my main concern is if the Monoid will match the proto3 specs related to default values, but I guess it'll.

Given that we are not running short on type-classes, and since the purposes of the default value in _proto3_ are very different from those of the `Monoid` type-class, it may be better to use a different type-class, say `ProtoDefault`, and add a special instance builder:
 
```Scala
def protoDefaultFromMonoid[A](implicit M: Monoid[A]): ProtoDefault[A] = 
  new ProtoDefault[A] { def default: A = M.zero }
```

noelmarkham:
@diesalbla You're right, the default proto values don't fully align to that of `Monoid`/`Empty`, so keeping it explicit and separate feels right.

Thanks for your input everyone, it's been really helpful.

PR #620, opened 2019-06-10
Introduce a typeclass for handling default values in protobuf messages

noelmarkham:
## What this does?
Introduce a `ProtoDefault` typeclass for handling default values, as required in #606.

Originally I was changing the call to `pbTo` from the `A` type to an `Option[A]` wrapped in a case class, but this was not playing nice with some particular option types when sent over the wire, similar to the tests found in `higherkindness.mu.rpc.protocol.RPCProtoProducts`, so I think it is fine here to wrap the call in a `Try`, and if that fails, return the default. I did make a start on changing pbdirect to return `Option[A]` rather than `A` for its parsing methods, but the approach I've taken feels to be the least invasive to fix the underlying bug.

The `ProtoDefault` has one field, `default`, and this can be derived from a `cats.Monoid` if there's one in scope. For the majority of defaults needed throughout the project, I have pulled in the Typelevel Kittens project to automatically derive a `Monoid` instance.

## Checklist

### DIFF
--- a/benchmarks/shared/src/main/scala/higherkindness/mu/rpc/benchmarks/shared/protocols/PersonServicePB.scala
+++ b/benchmarks/shared/src/main/scala/higherkindness/mu/rpc/benchmarks/shared/protocols/PersonServicePB.scala
@@ -21,6 +21,10 @@ package protocols
 import higherkindness.mu.rpc.benchmarks.shared.models._
 import higherkindness.mu.rpc.protocol._
 
+import cats.implicits._
+import cats.derived._
+import auto.monoid._
+
 @service(Protobuf)
 trait PersonServicePB[F[_]] {
 
--- a/modules/server/src/test/scala/higherkindness/mu/rpc/protocol/RPCMethodNameTests.scala
+++ b/modules/server/src/test/scala/higherkindness/mu/rpc/protocol/RPCMethodNameTests.scala
@@ -28,13 +28,10 @@ class RPCMethodNameTests extends RpcBaseTestSuite with BeforeAndAfterAll with Ch
 
   object RPCService {
 
-    implicit val protoDefaultRequest: ProtoDefault[Request] = new ProtoDefault[Request] {
-      override def default: Request = Request("")
-    }
-
-    implicit val protoDefaultResponse: ProtoDefault[Response] = new ProtoDefault[Response] {
-      override def default: Response = Response(0)
-    }
+    import cats.kernel.instances.string._
+    import cats.kernel.instances.int._
+    import cats.derived._
+    import auto.monoid._
 
     case class Request(s: String)
 
--- a/modules/server/src/test/scala/higherkindness/mu/rpc/protocol/RPCMethodNameTests.scala
+++ b/modules/server/src/test/scala/higherkindness/mu/rpc/protocol/RPCMethodNameTests.scala
@@ -28,6 +28,14 @@ class RPCMethodNameTests extends RpcBaseTestSuite with BeforeAndAfterAll with Ch
 
   object RPCService {
 
+    implicit val protoDefaultRequest: ProtoDefault[Request] = new ProtoDefault[Request] {
+      override def default: Request = Request("")
+    }
+
+    implicit val protoDefaultResponse: ProtoDefault[Response] = new ProtoDefault[Response] {
+      override def default: Response = Response(0)
+    }
+
     case class Request(s: String)
 
     case class Response(length: Int)
--- a/build.sbt
+++ b/build.sbt
@@ -12,7 +12,6 @@ lazy val common = project
   .in(file("modules/common"))
   .settings(moduleName := "mu-common")
   .settings(commonSettings)
-  .settings(kittensTODO)
 
 lazy val `internal-core` = project
   .in(file("modules/internal"))
@@ -20,7 +19,6 @@ lazy val `internal-core` = project
   .dependsOn(testing % "test->test")
   .settings(moduleName := "mu-rpc-internal-core")
   .settings(internalSettings)
-  .settings(kittensTODO)
 
 lazy val `internal-monix` = project
   .in(file("modules/internal/monix"))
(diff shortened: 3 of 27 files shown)


noelmarkham:
The plugin tests are currently failing as they're not generating a default. I'm not entirely sure of what's going to be the best way to deal with that, I'd be interested to hear some suggestions if you had any.

juanpedromoreno:
Hey @noelmarkham , thanks for your help here. To debug the issue of the scripted test (`srcProtoGenFromDirs `), I'd recommend the next steps:

* Publish locally the project based on your changes:
  * `sbt publishLocal`
  * `sbt idlgen-sbt/publishLocal`
* Then, go to the scripted test folder: https://github.com/higherkindness/mu/blob/master/modules/idlgen/plugin/src/sbt-test/sbt-mu-idlgen/srcGenFromJars/
* ... and load an sbt session in this way: `sbt -Dversion=<your version published in the first step>`
* Once loaded, you could run the commands inside the `test` file: https://github.com/higherkindness/mu/blob/master/modules/idlgen/plugin/src/sbt-test/sbt-mu-idlgen/srcGenFromJars/test ... in this case, running `publishLocal`

Please, let me know if it helps :)

noelmarkham:
Thanks - I think I'm more concerned about _how exactly_ it's going to generate a good enough `ProtoDefault` typeclass based on the types it's given, rather than how to debug the generated code itself, if that makes sense.

I'm wondering if automatic generation may be enough, I feel it could be if it has the right defaults for the underlying types, I'm taking a look at that.

modules/server/src/test/scala/higherkindness/mu/rpc/protocol/RPCMethodNameTests.scala
```diff
@@ -28,6 +28,11 @@ class RPCMethodNameTests extends RpcBaseTestSuite with BeforeAndAfterAll with Ch
 
   object RPCService {
 
+    import cats.kernel.instances.string._
```
diesalbla: Should these imports be added by the annotation within the generated class? Not asking it for this PR, but in case we need a follow-up.

benchmarks/shared/src/main/scala/higherkindness/mu/rpc/benchmarks/shared/protocols/PersonServicePB.scala
```diff
@@ -21,6 +21,10 @@ package protocols
 import higherkindness.mu.rpc.benchmarks.shared.models._
 import higherkindness.mu.rpc.protocol._
 
+import cats.implicits._
+import cats.derived._
+import auto.monoid._
+
```
rafaparadela: Do the users need to include these imports in all the protocols annotated with `@service(Protobuf)`?
I'm wondering why don't we just inject them in the macro, only in the case that `serializationType == Protobuf`.

noelmarkham: That sounds like a good idea

modules/server/src/test/scala/higherkindness/mu/rpc/protocol/RPCMethodNameTests.scala
noelmarkham: Yep, I think I should

benchmarks/shared/src/main/scala/higherkindness/mu/rpc/benchmarks/shared/protocols/PersonServicePB.scala
dzanot: So I [attempted this today](https://github.com/higherkindness/mu/compare/606-default-typeclass...dzanot:606-default-typeclass?expand=1#diff-90fc855999aa8ba5dc1873744eb5d339L29) and ran into some...very weird behavior. Moving the imports to the `service()` macro *mostly* worked, except in the case of `PersonServicePB`. I narrowed down the issue to the fact that its `PersonList` message contains a `List`. 

That is: `PersonServicePB` compiled with the kittens import in the `.scala` file but _not_ when added by the macro impl. All the other cases of `service(Protobuf)` were able to compile in both cases. 

So I wrote some sanity tests to prove that kittens can derive instances for such case classes, and discovered some other very strange behavior: To get the auto derivation to work and the test to compile I had to `include cats.implicits._` a _second_ time [as you can see here](https://github.com/higherkindness/mu/compare/606-default-typeclass...dzanot:606-default-typeclass?expand=1#diff-f749bcf748f83ddc00f394b3558820e9R90).

This discovery didn't solve the initial problem but it revealed to me some quirkiness that is beyond my understanding. Does anyone have any experience with multiple imports being necessary?

noelmarkham: I've never come across multiple imports being necessary, is this to do with it being in scope for the generated code but then not here unless it's imported?

dzanot: The test where multiple imports is "necessary" is independent of the generated code. It's testing the Monoid->ProtoDefault->Marshaller chain. As far as I can tell when I am importing the kittens implicits in the macro/generated code kittens isn't actually able to derive any monoid instances.

dzanot: FWIW I resolved the macro issues, duplicate instances of...instances were being imported because `pbd` inherited some https://github.com/dzanot/mu/commit/9ae27508501a5bf4993892be4e0e6de18079892f

L-Lavigne: Congrats on resolving this @dzanot! 🎉

noelmarkham: Oh nice! Thanks @dzanot!

fedefernandez:
@noelmarkham what is the status of this? Do you need someone to take a look into? Thanks

noelmarkham:
I've not had a chance to look at this since I've been with Avetta - I do
think Davis was taking a look though?

dzanot:
I was working on this, and was able to address some feedback here: https://github.com/higherkindness/mu/pull/628
But if I recall I was still unable to get tests to pass due to some bugs in how PBDirect handles oneof/coproducts.

cb372:
Closing this, as handling of default values and missing fields was added in v0.20.0.

Issue #631, opened 2019-07-29
Replace macros with code generation


L-Lavigne:
Due to ever-changing standards and poor IDE support for Scala macros, we should look into potential replacements for those as we migrate Mu towards Scala 2.13 and beyond.

(For reference, the current implementation is [here in serviceImpl.scala](https://github.com/higherkindness/mu/blob/635abd0bfb93ce18f3419682639327b8786ffcc7/modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala).)

It has proven difficult in the past to predict which direction Scala macros will go. For example, here in Mu we [migrated from `scalameta.paradise` to `scalamacros.paradise`](https://github.com/higherkindness/mu/issues/291), only to then see the _former_ library [integrated into the Scala compiler](https://github.com/scala/scala/pull/6606) in [2.13.0-M4](https://github.com/scala/scala/releases/tag/v2.13.0-M4).

Although this compiler integration gives hope for greater standardization and stability in the future, there are still significant downsides to using macros in this project. Foremost is the opacity of the generated code and its poor support by IDEs such as IntelliJ IDEA. Generated client and server types are generally unrecognized (marked red) by the IDE, and visual macro expansion toggles work unreliably if at all.

Using some form of code generation, yielding sources plainly visible to the developer, could alleviate these concerns.

Right now I'm looking at using a compiler plugin and the `scala.tools.nsc._` tooling which allows re-using most of the existing quasiquotes implementation without involving macros. The main downside so far is that since we need one compilation process for generating the sources and another one to process them, we'd require the user to set up their services in a separate sbt subproject from the program that calls them. I believe this can be made a smooth process through good examples and tutorials, but I'm happy to hear different opinions on this.

I made a non-Mu POC with basic codegen that works with Scala 2.13 (Mu's dependencies are not yet 2.13-ready, hence the need for a side project). I'm going to port the code to a Mu branch to confirm it can work in 2.12 within the project's context. I'll provide a draft PR for this shortly.

Comments and suggestions welcome!

fedefernandez:
@L-Lavigne this looks good to me. This is an idea that we've been thinking about since some time ago. Thanks for trying to address it.

As a side note, I think it'd be very useful for mu-kotlin, in order to keep both libraries synchronized.

L-Lavigne:
POC PR (work in progress): #632

PR #632, opened 2019-08-03
Replace macros with codegen - POC/WIP

L-Lavigne:
This PR introduces a work-in-progress compiler plugin (not an sbt plugin) to generate "managed" sources from the `@service` definitions (here `@serviceCodegen`, to preserve the originals) without using macros but by reusing most of the existing macro implementation. Note that this requires a 2-subproject setup so that the app's compilation can see the sources generated by the protocol's compilation. This is shown in `example/codegen`.

Please refer to #631 for the rationale behind this and the pros & cons of this approach.

How it works: Compiling `example/codegen/app` should trigger compilation of `example/codegen/protocol` which will generate the object in the protocol module's `target/scala-2.12/src_managed` for the app module to refer to. The generator is in the `common` module's `ServiceGenerator.scala` and basically wraps a modified copy of the `internal` module's `serviceImpl` inside a compiler plugin instead of a macro.

Note the following current limitations and remaining TODOs:
- The compiler plugin resides in the `common` project (where the protocol keywords are) instead of depending on it, because of the limitations described in https://github.com/sbt/sbt/issues/2255. We should move it to a different module and implement a workaround based on sbt-assembly.
- The new annotation is temporarily called `@serviceCodegen` to preserve the existing `@service` macro's functionality; the goal if we adopt this approach is to replace the `@service` implementation.
- The generated object is simply called `${service}Object` and contains everything the macro currently generates in the service companion; we'll want to split this into client and server classes for better readability.
- The generated object is in a hardcoded `example/codegen` package to match the example modules, but we'll change it to the actual service's package.
- The generated object doesn't have the imports of the service trait; we'll need to add them to the generated source.
- The example module doesn't do anything with the generated object except confirm that it can see it; we'll want to add interesting service methods and call them.

Comments, questions and ideas welcome!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -12,6 +12,7 @@ lazy val common = project
   .in(file("modules/common"))
   .settings(moduleName := "mu-common")
   .settings(commonSettings)
+  .settings(codegenSettings)
 
 lazy val `internal-core` = project
   .in(file("modules/internal"))
@@ -466,6 +467,35 @@ lazy val `example-todolist-client` = project
   .settings(moduleName := "mu-rpc-example-todolist-client")
   .settings(exampleTodolistCommonSettings)
 
+/////////////////
+//// CODEGEN ////
+/////////////////
+
+lazy val `example-codegen-protocol` = project
+  .in(file("modules/examples/codegen/protocol"))
+  .dependsOn(`channel`, `common` % Configurations.CompilerPlugin)
+  .settings(coverageEnabled := false)
+  .settings(noPublishSettings)
+  .settings(moduleName := "mu-rpc-example-codegen-protocol")
+  .settings(
+    scalacOptions ++= Seq(
+      // recompile whenever plugin changes (in IDEA this might require using sbt shell for imports & builds)
+      "-Jdummy" + (`common`/Compile/packageBin).value.lastModified,
+      "-P:service-generator:outputDir=" + (Compile/sourceManaged).value
+    ),
+    autoCompilerPlugins := true,
+  )
+
+lazy val `example-codegen-app` = project
+  .in(file("modules/examples/codegen/app"))
+  .dependsOn(`example-codegen-protocol`)
+  .settings(coverageEnabled := false)
+  .settings(noPublishSettings)
+  .settings(moduleName := "mu-rpc-example-codegen-app")
+  .settings(
+    Compile/sourceGenerators += Def.task(((`example-codegen-protocol`/Compile/sourceManaged).value ** "*.scala").get)
+  )
+
 /////////////////////
 //// MARSHALLERS ////
 /////////////////////
@@ -531,6 +561,8 @@ lazy val allModules: Seq[ProjectReference] = Seq(
   `example-todolist-runtime`,
   `example-todolist-server`,
   `example-todolist-client`,
+  `example-codegen-protocol`,
+  `example-codegen-app`,
   seed,
   `benchmarks-vprev`,
   `benchmarks-vnext`,
--- a/modules/common/src/main/scala/higherkindness/mu/rpc/codegen/ServiceGenerator.scala
+++ b/modules/common/src/main/scala/higherkindness/mu/rpc/codegen/ServiceGenerator.scala
@@ -1,0 +1,720 @@
+/*
+ * Copyright 2017-2019 47 Degrees, LLC. <http://www.47deg.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package higherkindness.mu.rpc.codegen
+
+import higherkindness.mu.rpc.protocol._
+import java.io._
+import scala.tools.nsc._
+import scala.tools.nsc.plugins._
+
+// $COVERAGE-OFF$
+
+object ServiceGenerator {
+  val OutputDirOption = "outputDir"
+}
+
+class ServiceGenerator(val global: Global) extends Plugin {
+  import ServiceGenerator._
+
+  val name        = "service-generator"
+  val description = "Generates client and server sources from @service annotations"
+
+  override val optionsHelp: Option[String] = Some(
+    s"$OutputDirOption: output directory for generated client and server sources")
+
+  private val parsedOptions: Map[String, String] = options.map { option =>
+    val parsedOption = option.split('=')
+    require(parsedOption.length == 2, s"Invalid option syntax: $option")
+    parsedOption.head -> parsedOption.last
+  }.toMap
+
+  private val outputDir = new File(option(OutputDirOption))
+
+  lazy val components: List[PluginComponent] = List(
+    new ServiceGeneratorComponent(global, name, outputDir))
+
+  override def init(options: List[String], error: String => Unit): Boolean = true
+
+  private def option(optionName: String): String =
+    parsedOptions.getOrElse(
+      optionName,
+      throw new IllegalArgumentException(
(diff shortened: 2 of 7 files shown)


build.sbt
```diff
@@ -466,6 +467,35 @@ lazy val `example-todolist-client` = project
...
+lazy val `example-codegen-protocol` = project
+  .in(file("modules/examples/codegen/protocol"))
+  .dependsOn(`channel`, `common` % Configurations.CompilerPlugin)
+  .settings(coverageEnabled := false)
+  .settings(noPublishSettings)
+  .settings(moduleName := "mu-rpc-example-codegen-protocol")
+  .settings(
+    scalacOptions ++= Seq(
+      // recompile whenever plugin changes (in IDEA this might require using sbt shell for imports & builds)
+      "-Jdummy" + (`common`/Compile/packageBin).value.lastModified,
```
juanpedromoreno:
```suggestion
      "-Jdummy" + (`common` / Compile / packageBin).value.lastModified,
```

build.sbt
```diff
@@ -466,6 +467,35 @@ lazy val `example-todolist-client` = project
...
+      "-P:service-generator:outputDir=" + (Compile/sourceManaged).value
```
juanpedromoreno:
```suggestion
      "-P:service-generator:outputDir=" + (Compile / sourceManaged).value
```

build.sbt
```diff
@@ -466,6 +467,35 @@ lazy val `example-todolist-client` = project
...
+  )
+
+lazy val `example-codegen-app` = project
+  .in(file("modules/examples/codegen/app"))
+  .dependsOn(`example-codegen-protocol`)
+  .settings(coverageEnabled := false)
+  .settings(noPublishSettings)
+  .settings(moduleName := "mu-rpc-example-codegen-app")
+  .settings(
+    Compile/sourceGenerators += Def.task(((`example-codegen-protocol`/Compile/sourceManaged).value ** "*.scala").get)
```
juanpedromoreno:
```suggestion
    Compile / sourceGenerators += Def.task(
      ((`example-codegen-protocol` / Compile / sourceManaged).value ** "*.scala").get)
```

modules/common/src/main/scala/higherkindness/mu/rpc/codegen/ServiceGenerator.scala
```diff
@@ -0,0 +1,720 @@
...
+}
+
+class ServiceGeneratorComponent(val global: Global, val phaseName: String, outputDir: File)
+    extends PluginComponent {
+  import global._
+
+  private lazy val serviceFilter = new FilterTreeTraverser({
+    case md: MemberDef =>
+      findAnnotation(md.mods, "serviceCodeGen").isDefined // TODO: rename to "service"
+    case x => false
```
juanpedromoreno:
```suggestion
    case _ => false
```

modules/common/src/main/scala/higherkindness/mu/rpc/codegen/ServiceGenerator.scala
```diff
@@ -0,0 +1,720 @@
...
+  lazy val components: List[PluginComponent] = List(
+    new ServiceGeneratorComponent(global, name, outputDir))
+
+  override def init(options: List[String], error: String => Unit): Boolean = true
+
+  private def option(optionName: String): String =
+    parsedOptions.getOrElse(
+      optionName,
+      throw new IllegalArgumentException(
+        s"Plugin $name missing option $optionName.\n${optionsHelp.get}"))
```
juanpedromoreno: Could we avoid this `Option.get`?

modules/common/src/main/scala/higherkindness/mu/rpc/codegen/ServiceGenerator.scala
```diff
@@ -0,0 +1,720 @@
...
+            q"_root_.io.grpc.stub.ServerCalls.asyncUnaryCall(_root_.higherkindness.mu.rpc.internal.server.unaryCalls.unaryMethod(algebra.$name, $compressionOption))"
+          case _ =>
+            sys.error(
+              s"Unable to define a handler for the streaming type $streamingType and $prevalentStreamingTarget for the method $name in the service $serviceName")
+        }
+        q"($methodDescriptorName, $handler)"
+      }
+    }
+
+    case class HttpOperation(operation: Operation) {
```
juanpedromoreno: > we'll want to split this into client and server classes for better readability.

Would it make sense to split the Http part too?

modules/examples/codegen/protocol/src/main/scala/example/codegen/ExampleService.scala
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+package example.codegen
+
+import higherkindness.mu.rpc.protocol._
+
+@message case class Hello(words: String)
+
+@serviceCodeGen(Protobuf) trait ExampleService[F[_]] {
```
juanpedromoreno: I wonder how the code generation would behave when we would be generating code in two different stages:

1) From IDL to Scala code (scala files with `@serviceCodeGen`) (**sbt-mu-idl-gen**)
2) `@serviceCodeGen` to the expanded version of the Scala Code (**scala compiler plugin**)

modules/common/src/main/scala/higherkindness/mu/rpc/codegen/ServiceGenerator.scala
L-Lavigne: Absolutely! Currently we have a monolithic straight-up copy of the current macro output but there's lots of improvements we can make so it becomes more user-friendly.

Also moving off macros might help resolve the issue that led us to put the HTTP stuff in there in the first place, where we couldn't get two processors for the same macro applied from different classes.

L-Lavigne:
> Probably, we'll want to provide the compiler plugin automatically when you are including the `sbt-mu-idl-gen` sbt plugin in the project. Would it make sense?

The compiler plugin processes the `@service` annotation and as such would be part of the core functionality of Mu just as the macro is today, so its module should always be imported by developers whether or not they're using IDL generation.

modules/examples/codegen/protocol/src/main/scala/example/codegen/ExampleService.scala
L-Lavigne: I haven't tried it yet but I suppose the SBT plugin generates the `@service`-annotated sources from IDL, and then the compiler plugin generates the service infrastructure from those sources afterwards. 

We could explore combining these two steps into one, however I don't think that would work since the developer needs the intermediate `@service`-annotated traits in order to implement their methods.

L-Lavigne:
Note that the CI build is failing because the generated sources are either not created by the example protocol or not seen by the example app. It _Works on my Machine_ (™) so I'll investigate what happened there.

rafaparadela:
@L-Lavigne CI is failing because you need to `project example-codegen-app` and `compile`.

L-Lavigne:
> @L-Lavigne CI is failing because you need to `project example-codegen-app` and `compile`.

That should be triggered by the root project's compile though, and it should trigger compilation of `example-codegen-protocol` which uses the plugin to generate the service object. I do see this project compiling here: https://travis-ci.org/higherkindness/mu/jobs/567183204#L968, and it seems like it's either not using the plugin, or the `example-codegen-app` compilation isn't adding the protocol's generated sources.

Still not sure why this is the case. Locally it works although I do recall needing a full project re-import in IDEA at some point. I don't see previous build state being much of an issue for CI though.

ivanov-ia:
This doesn't compile for me when using sbt console and running `example-codegen-app/compile` for the first time or after `clean`. For the second time it compiles without errors.

The reason is that the task at https://github.com/higherkindness/mu/pull/632/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR496 is evaluated before the plugin generates ExampleServiceObject.scala. You can see this, if you use:
```
Compile/sourceGenerators += Def.task {
      val files = ((`example-codegen-protocol` / Compile / sourceManaged).value ** "*.scala").get
      println("FILES: " + files.mkString(", "))
      files
}
```
The list of files is empty and is printed before the compilation of `example-codegen-protocol`.

The example works fine this way:
```
Compile/sourceGenerators += Def.sequential(
    `example-codegen-protocol` / Compile / compile,
    Def.task(((`example-codegen-protocol` / Compile / sourceManaged).value ** "*.scala").get)
)
```
But it doesn't seem to be a good solution.

cb372:
Closing. We did end up going with srcgen, but not quite like this.

PR #642, opened 2019-09-19
Upgrades Project

juanpedromoreno:
## What this does?

This PR bumps up all the project dependencies to the latest.

Additionally removes the support for Scala 2.11, as some projects like Circe or http4s in their latest versions did (and the `http` module cannot cross-compile).

In futures PRs we will need to add support for Scala 2.13.

## Checklist


fedefernandez:
@juanpedromoreno It looks like it has some binary incompatibilities related with fs2

juanpedromoreno:
@fedefernandez I've removed the support for Scala 2.11, as some projects like Circe or http4s in their latest versions did (and the http module cannot cross-compile).

juanpedromoreno:
It's now ready for review 👀 . Thanks!

PR #726, opened 2019-12-18
Upgrade to latest snapshots of pbdirect and skeuomorph

cb372:
This version of pbdirect requires all fields to be annotated with a `@pbIndex` and only allows messages to be encoded, not individual fields.

This version of skeuomorph generates code that includes those annotations.

It's a huge diff but it's mostly just me going around and adding the annotation everywhere.

Note: the CI build will fail because it's relying on locally built snapshots.

## What this does?

## Checklist


modules/internal/src/main/scala/higherkindness/mu/rpc/internal/encoders/pbd.scala
```diff
@@ -30,7 +30,7 @@ object pbd extends OptionInstances with ListInstances {
 
   import pbdirect._
 
-  implicit def defaultDirectPBMarshallers[A: PBWriter: PBReader]: Marshaller[A] =
+  implicit def defaultDirectPBMarshallers[A: PBMessageWriter: PBReader]: Marshaller[A] =
```
cb372: This change is quite important. It means it's no longer in valid in Mu to create an RPC endpoint that takes a single field (e.g. a `String`) as a request or return one as a response. You have to use a proper message type (e.g. `case class MyRequest(@pbIndex(1) id: String)` for both requests and responses.

@juanpedromoreno Do you think this is a reasonable change? I found quite a few places in the tests where we were using things like `LocalDateTime` or `BigDecimal` as the RPC request type. I had to delete those tests because they don't compile now.

The only reason for this change is that I did some refactoring in my pbdirect PR, splitting out `PBWriter` into `PBMessageWriter` and `PBFieldWriter`. If necessary I can revert that refactoring and put everything back to how it was before.

modules/kafka/src/main/scala/higherkindness/mu/rpc/kafka/kafkaManagementService.scala
```diff
@@ -291,43 +316,46 @@ object kafkaManagementService {
...
-  final case class TopicListings(listings: List[TopicListing])
+  final case class TopicListings(@pbIndex(1) listings: List[TopicListing])
 
   @service(Protobuf)
   trait KafkaManagement[F[_]] {
     def alterConfigs(acr: AlterConfigsRequest): F[Unit]
     def createPartitions(cpr: CreatePartitionsRequest): F[Unit]
     def createTopic(ctr: CreateTopicRequest): F[Unit]
     def createTopics(ctrs: CreateTopicRequests): F[Unit]
-    def deleteTopic(t: String): F[Unit]
```
cb372: Here's an example of an RPC request that's no longer allowed

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/encoders/pbd.scala
juanpedromoreno: I think it's definitely a reasonable change 👍

modules/kafka/src/main/scala/higherkindness/mu/rpc/kafka/kafkaManagementService.scala
juanpedromoreno: Yeah, it makes sense if we need to wrap it up into a case class

PR #746, opened 2020-01-15
Scalameta codegen

cb372:
## What this does?

Bump the skeuomorph dependency and use its new scalameta-based printer for source code generation.

## Checklist


cb372:
The coverage went down but it's pretty much impossible to test the error-handling branch. An error in this case basically means a bug in skeuomorph.

PR #769, opened 2020-01-28
Updates to scalatest 3.1.0

juanpedromoreno:
## What this does?

This PR updates to Scalatest 3.1.0, adding the new `scalatestplus-scalacheck` dependency, as it's needed from now on.

## Checklist

### DIFF
--- a/project/ProjectPlugin.scala
+++ b/project/ProjectPlugin.scala
@@ -320,7 +320,10 @@ object ProjectPlugin extends AutoPlugin {
     )
 
     lazy val docsSettings: Seq[Def.Setting[_]] = Seq(
-      libraryDependencies += %%("scalatest", V.scalatest),
+      libraryDependencies ++= Seq(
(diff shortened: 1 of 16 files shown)


project/ProjectPlugin.scala
```diff
@@ -58,7 +58,8 @@ object ProjectPlugin extends AutoPlugin {
       val scalacheck: String         = "1.14.3"
       val scalacheckToolbox: String  = "0.3.1"
       val scalamock: String          = "4.4.0"
-      val scalatest: String          = "3.0.8"
+      val scalatest: String          = "3.1.0"
+      val scalatestplusScheck        = "3.1.0.0-RC2"
```
fedefernandez: The dependency is not aligned with scalatest? Should we add the type to make it consistent?

juanpedromoreno: No, it's not. From 3.1.0 they were separated under its own versioning.

PR #772, opened 2020-01-29
Adds a prometheus builder for full metrics

### LINKED ISSUE #635: Prometheus server latency metrics no longer provide labels per service and method
We recently upgraded a service from `v0.14.1` to `v0.18.4`. In the old version there was a way to [opt in](https://github.com/higherkindness/mu/blob/v0.14.1/modules/prometheus/server/src/main/scala/ServerMetrics.scala#L55) to latency metrics with [labels per service and method](https://github.com/higherkindness/mu/blob/v0.14.1/modules/prometheus/server/src/main/scala/ServerMetricsForMethod.scala#L47). I may be missing something, but I'm not seeing a mechanism provided that would allow for the same sort of per-endpoint latency metrics. If that is the case, is there maybe a reason those were removed that I might be missing? And if there is no reason, would anyone be opposed to adding something back in? I would be happy to submit a PR if there are no objections to the idea.

fedefernandez:
## What this does?

Fixes #635

This adds a new builder for construct a prometheus metrics with the service and method names included in the total time histogram.

Since histograms could have some performance problems when the granularity of the labels increase, we can remove the headers histogram from the new builder (which is not providing too much value due to the number of labels). The user could always add their own metrics.

This also provides a sample about how to tune the metric builder.

## Checklist

### DIFF
--- a/modules/metrics/prometheus/src/main/scala/higherkindness/mu/rpc/prometheus/PrometheusMetrics.scala
+++ b/modules/metrics/prometheus/src/main/scala/higherkindness/mu/rpc/prometheus/PrometheusMetrics.scala
@@ -33,9 +33,9 @@ import io.prometheus.client._
  * {prefix}_messages_sent{labels=classifier,service,method} - Counter
  * {prefix}_messages_received{labels=classifier,service,method} - Counter
  * {prefix}_calls_header{labels=classifier} - Histogram
- * {prefix}_calls_total{labels=classifier,method,status} - Histogram
+ * {prefix}_calls_total{labels=classifier,methodType,status} - Histogram
  *
- * `method` can be one of the following:
+ * `methodType` can be one of the following:
  *    - "unary"
  *    - "client-streaming"
  *    - "server-streaming"
@@ -54,6 +54,12 @@ import io.prometheus.client._
  *    - "unknown-status"
  *    - "unreachable-error"
  *
+ * The `buildFullTotal` method contains the following metrics:
+ *
+ * {prefix}_active_calls{labels=classifier} - Counter
+ * {prefix}_messages_sent{labels=classifier,service,method} - Counter
+ * {prefix}_messages_received{labels=classifier,service,method} - Counter
+ * {prefix}_calls_total{labels=classifier,service,method,status} - Histogram
  */
 object PrometheusMetrics {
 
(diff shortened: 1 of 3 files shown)


modules/metrics/prometheus/src/main/scala/higherkindness/mu/rpc/prometheus/PrometheusMetrics.scala
```diff
@@ -33,9 +33,9 @@ import io.prometheus.client._
  * {prefix}_messages_sent{labels=classifier,service,method} - Counter
  * {prefix}_messages_received{labels=classifier,service,method} - Counter
  * {prefix}_calls_header{labels=classifier} - Histogram
- * {prefix}_calls_total{labels=classifier,method,status} - Histogram
+ * {prefix}_calls_total{labels=classifier,methodType,status} - Histogram
```
fedefernandez: This was confusing since we had in the messages sent/received metrics the same label name but with different values.

modules/metrics/prometheus/src/main/scala/higherkindness/mu/rpc/prometheus/PrometheusMetrics.scala
```diff
@@ -54,6 +54,12 @@ import io.prometheus.client._
  *    - "unknown-status"
  *    - "unreachable-error"
  *
+ * The `buildFullTotal` method contains the following metrics:
```
juanpedromoreno: Maybe we can name it as `buildAllTotalMetrics`, would it make sense?

modules/metrics/prometheus/src/test/scala/higherkindness/mu/rpc/prometheus/PrometheusMetricsTests.scala
```diff
@@ -185,4 +185,31 @@ class PrometheusMetricsTests extends Properties("PrometheusMetrics") {
...
+      (methodInfo: GrpcMethodInfo, numberOfCalls: Int, status: Status, elapsed: Int) =>
+        val registry   = new CollectorRegistry()
+        val metricName = s"${prefix}_calls_total"
+
+        (for {
+          metrics <- PrometheusMetrics.buildFullTotal[IO](registry, prefix)
+          op1 <- (1 to numberOfCalls).toList
+            .map(_ => metrics.recordTotalTime(methodInfo, status, elapsed.toLong, Some(classifier)))
+            .sequence_
+            .map { _ =>
```
juanpedromoreno: Maybe we can simplify it with `traverse`:

```scala

          op1 <- (1 to numberOfCalls).toList
            .traverse(_ => metrics.recordTotalTime(methodInfo, status, elapsed.toLong, Some(classifier)))
            .map { _ =>
              checkMetrics(
```

modules/metrics/prometheus/src/main/scala/higherkindness/mu/rpc/prometheus/PrometheusMetrics.scala
fedefernandez: I think "full" is more appropriated, although potentially improved. I'm referring here to a more complete metrics than something that including more metrics. wdyt?

PR #773, opened 2020-01-29
Introduce sum types for `idlType` and `serializationType`

cb372:
## Checklist


cb372:
Forgot to update the docs (and then accidentally opened the PR anyway). I'll do that now.

PR #798, opened 2020-02-21
Remove the `@message` annotation

cb372:
It was a relic from the IDL generation feature which has been removed.

## What this does?

## Checklist


modules/common/src/main/scala/higherkindness/mu/rpc/protocol/protocol.scala
```diff
@@ -38,10 +38,8 @@ sealed abstract class MethodNameStyle extends Product with Serializable
 case object Unchanged                 extends MethodNameStyle
 case object Capitalize                extends MethodNameStyle
 
-class message                          extends StaticAnnotation
 class option(name: String, value: Any) extends StaticAnnotation
 class outputPackage(value: String)     extends StaticAnnotation
 class outputName(value: String)        extends StaticAnnotation
```
cb372: I think these annotations can also go, but I wanted to keep the PR small

PR #807, opened 2020-02-24
Cross compile code with Scala 2.13.x

### LINKED ISSUE #787: Cross compilation with scala 2.13
Could we add cross-compilation with 2.13? AFAIK this is the only library we use without being published for 2.13 on the cats ecosystem and it's blocking some services to be upgraded. 
Is there any blocker to achieving this?

juanpedromoreno:
## What this does?

Closes #787 

## Checklist


juanpedromoreno:
Some of current test issues:

```bash
[error] 	higherkindness.mu.rpc.http.GreeterDerivedRestTests
[error] 	higherkindness.mu.rpc.http.GreeterRestTests
[error] (http / Test / test) sbt.TestsFailedException: Tests unsuccessful
```

.travis.yml
```diff
@@ -66,6 +82,7 @@ jobs:
             sbt release;
           fi
     - stage: microsite
+      scala: 2.13.1
```
AdrianRaFo: I suppose this is required but we are publishing for both right?

build.sbt
```diff
@@ -585,6 +593,18 @@ lazy val allModules: Seq[ProjectReference] = Seq(
   http,
   kafka,
   `marshallers-jodatime`,
+  `legacy-avro-decimal-compat-model`,
+  `legacy-avro-decimal-compat-protocol`,
+  `legacy-avro-decimal-compat-encoders`,
+  `health-check-unary`,
+  `example-health-client`,
+  `example-health-server-monix`,
+  `example-health-server-fs2`
```
AdrianRaFo: Why is this the only example on the core?

project/ProjectPlugin.scala
```diff
@@ -47,13 +45,15 @@ object ProjectPlugin extends AutoPlugin {
       val logback: String             = "1.2.3"
       val monix: String               = "3.1.0"
       val monocle: String             = "2.0.2"
+      val lastReleasedV               = "0.20.1"
```
AdrianRaFo:
```suggestion
      val lastReleased               = "0.20.1"
```

build.sbt
```diff
@@ -585,43 +588,55 @@ lazy val allModules: Seq[ProjectReference] = Seq(
   http,
   kafka,
   `marshallers-jodatime`,
+  `legacy-avro-decimal-compat-model`,
+  `legacy-avro-decimal-compat-protocol`,
+  `legacy-avro-decimal-compat-encoders`,
+  `health-check-unary`,
+  `example-health-client`,
+  `example-health-server-monix`,
+  `example-health-server-fs2`,
```
cb372: How do we decide which modules are "core" and which are not? Example projects don't seem very core to me.

Is `coreModules` actually the list of projects that the docs depend on? If so, let's rename it.

build.sbt
```diff
@@ -585,43 +588,55 @@ lazy val allModules: Seq[ProjectReference] = Seq(
...
   `example-routeguide-protocol`,
   `example-routeguide-common`,
   `example-routeguide-runtime`,
   `example-routeguide-server`,
   `example-routeguide-client`,
+  seed
```
cb372: `seed` is also an example project, right? Shall we rename it to `example-seed` for consistency?

build.sbt
```diff
@@ -585,43 +588,55 @@ lazy val allModules: Seq[ProjectReference] = Seq(
...
+lazy val coreModulesDeps: Seq[ClasspathDependency] =
+  coreModules.map(ClasspathDependency(_, None))
 
 lazy val root = project
   .in(file("."))
   .settings(name := "mu-scala")
   .settings(noPublishSettings)
-  .aggregate(allModules: _*)
-  .dependsOn(allModulesDeps: _*)
+  .aggregate(coreModules: _*)
```
cb372: Oh, root doesn't aggregate all the example projects any more? I'm not sure that's a good idea. I can imagine myself doing a refactoring, running `compile` or `test` and thinking everything's ok, but actually my refactoring has broken half the example projects.

project/ProjectPlugin.scala
```diff
@@ -358,13 +403,12 @@ object ProjectPlugin extends AutoPlugin {
         organizationHomePage = url("http://47deg.com"),
         organizationEmail = "hello@47deg.com"
       ),
-      scalaVersion := V.scala,
-      crossScalaVersions := Seq(V.scala),
-      scalacOptions ++= customScalacOptions,
+      scalaVersion := V.scala213,
+      crossScalaVersions := Seq(V.scala212, V.scala213),
+      scalacOptions ~= (_ filterNot Set("-Xfuture", "-Xfatal-warnings").contains),
```
cb372: Why are we disabling `fatal-warnings`?

cb372: Also you can write this more simply as `scalacOptions --= Seq("-Xfuture", "-Xfatal-warnings")`

cb372:
I'm a little confused about the rearrangement of the modules, but overall this looks great! Thank you for tackling it.

project/ProjectPlugin.scala
juanpedromoreno: > Why are we disabling fatal-warnings?

Current code won't compile as it is. If we want to enable it, I propose to do it in a separate PR.

> Also you can write this more simply as scalacOptions --= Seq("-Xfuture", "-Xfatal-warnings")

👍 https://github.com/higherkindness/mu-scala/pull/807/commits/263c3a5fc8426d3ef42772e41a9a9e9c0bb8e5b6

build.sbt
juanpedromoreno: I think the amount of modules is currently insane. My proposal would be to move all the examples to its own repository.

```scala
lazy val nonCrossedScalaVersionModules: Seq[ProjectReference] = Seq(
  `benchmarks-vprev`,
  `benchmarks-vnext`,
  `example-todolist-protocol`,
  `example-todolist-runtime`,
  `example-todolist-server`,
  `example-todolist-client`
)
```

Currently, the `root` module cannot aggregate the modules above because they cannot be cross-compiled with Scala 2.13.x. That's why I split the aggregation into two different modules and added to Travis CI the proper check to make sure nothing is broken before going to master.

build.sbt
juanpedromoreno: Done in https://github.com/higherkindness/mu-scala/commit/263c3a5fc8426d3ef42772e41a9a9e9c0bb8e5b6

build.sbt
juanpedromoreno: I've renamed the `others` by something more explicit. Basically, the modules that are not part of the `coreModules` are those that cannot be cross-compiled to Scala 2.13.x.

build.sbt
juanpedromoreno: Proposal filed at https://github.com/higherkindness/mu-scala/issues/817

cb372: >I think the amount of modules is currently insane.

Agreed. It makes the CI painfully slow, and navigating the code is really difficult with the library and the examples all mixed together in the same repo. It makes the examples less discoverable as well.

PR #813, opened 2020-03-02
Automatically register muSrcGen as a sourceGenerator

cb372:
I'm working on the documentation at the moment, and noticed this while I was going through a tutorial.

It doesn't really make sense *not* to add the task to the list of sourceGenerators, as the list of generated source files would not be added to `managedSources`. That means even if you ran `muSrcGen` manually and then ran `compile`, the generated source files would not be compiled.

Updated the plugin to automatically register the `muSrcGen` task as a sourceGenerator, and updated the misleading guidance in the documentation.

## What this does?

## Checklist

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -206,7 +206,7 @@ lazy val kafka = project
   .settings(kafkaSettings)
 
 ////////////////
-//// IDLGEN ////
+//// SRCGEN ////
 ////////////////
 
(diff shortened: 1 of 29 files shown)


cb372:
@BenFradet Yes, the tests have started failing now that they are actually testing something 😄 I'll fix them.

PR #820, opened 2020-03-05
Docs refresh

cb372:
Pretty much a complete rewrite of the microsite.

Inspired by [this excellent blogpost](https://www.divio.com/blog/documentation/) and [the gRPC docs](https://grpc.io/docs/), I've separated the documentation into a few categories:

* Getting started
    * This is for new users who just want to get started with minimal effort and see Mu-Scala doing something. Good for "just show me some code"-style learners.
    * Based on a new giter8 template: https://github.com/higherkindness/mu-scala.g8
* Tutorials
    * Lessons for Mu-Scala beginners. Narrative style, mix of explaining concepts and achieving practical steps.
* How-To guides
    * Shorter introductions of a particular Mu-Scala feature. Designed for users already familiar with Mu-Scala.
* Reference
    * Boring but important stuff, e.g. sbt settings, or the list of modules and artifacts. No narrative, just facts.

Overall goals:

* Make it easier for new users to get started with Mu-Scala
* Help both new and experienced users to find the information they need, in a suitable format
* Introduce information when it is most relevant
* Make it clear that RPC is just one feature of Mu, not the whole thing
* Make it easier to extend the documentation in the future, thanks to the clearer structure

This is still a WIP. I have to write some more content but the navigation structure is in place, so I'm ready to start receiving feedback.

The diff will be really hard to review, so I've published the site to my fork for people to read through: https://cb372.github.io/mu-scala/

I'd appreciate a review from @serras or @kutyel as you've been writing a lot of docs recently, and we should try to stay aligned on stuff like this.

If anyone wants to try going through the tutorials, that would be great. You'll need to clone this repo and publish a local snapshot first: `sbt publishLocal srcgen-sbt/publishLocal`


docs/src/main/docs/tutorials/grpc-server-client.md
```diff
@@ -0,0 +1,172 @@
...
+  with Protobuf](grpc-with-protobuf) tutorial
+
+Mu supports both Protobuf and Avro. For the purposes of this tutorial we will
+assume you are using Protobuf, but it's possible to follow the tutorial even if
+you are using Avro.
+
+## Service definition
+
+If you have followed one of the previous tutorials, you should already have a
+service definition that looks like this:
```
juanpedromoreno: Not a big deal, but maybe mentioning that this particular code is related to protobuf?

cb372:
@calvellido Thanks for the review!

I agree, but we already have a few nested menus, and sbt-microsites does not currently support more than one level of nested menu. I tried patching sbt-microsites to support 2 levels of nesting, but it looked kinda weird so I gave up on it.

Also having a single menu for the whole site would make the menu really long. The menu would be longer than the screen, and I'm not sure how well sbt-microsites's CSS supports that.

israelperezglez:
One solution would be to add a breadcrumb, that way the user knows what level of navigation he is in at all times.

But the best solution is to add navigation levels in the 

;)

serras:
Very nice content! I have just one suggestion: cut a bit the first page, because (at least) I need to scroll to get to the *Getting Started* section. Maybe start with a few lines describing Mu, then point people to guides and tutorials, and then include the list of features?

docs/src/main/docs/tutorials/testing-rpc-service.md
```diff
@@ -0,0 +1,204 @@
...
+
+We need to provide a couple of implicits to make that work:
+
+```scala mdoc:silent
+import cats.effect.{IO, ContextShift, Timer}
+import scala.concurrent.ExecutionContext.Implicits
+
+trait CatsEffectImplicits {
+
+  val EC: scala.concurrent.ExecutionContext = Implicits.global
```
BenFradet: :art:, since you're not making use of the ec implicitly, you can do:

```suggestion
import scala.concurrent.ExecutionContext

trait CatsEffectImplicits {

  val EC: ExecutionContext = ExecutionContext.global
```

docs/src/main/docs/tutorials/testing-rpc-service.md
```diff
@@ -0,0 +1,204 @@
...
+* are new to Mu-Scala
+* have some understanding of Protobuf and `.proto` file syntax
+* have read the [Getting Started guide](../getting-started)
+* have followed the [gRPC server and client tutorial](grpc-server-client)
+
+## Service definition
+
+Let's use the following service definition. (For an explanation of how to create
+service definition, check out the [RPC service definition with Protobuf
+tutorial](service-definition/protobuf).)
```
BenFradet: it might be worth mentioning that while this guide is using Protobuf, the notions covered are also useful for Avro users, wdyt?

BenFradet: maybe in the section above

docs/src/main/docs/tutorials/testing-rpc-service.md
```diff
@@ -0,0 +1,204 @@
...
+   */
+  val clientResource: Resource[IO, Greeter[IO]] = for {
+    sc        <- withServerChannel(Greeter.bindService[IO])
+    clientRes <- Greeter.clientFromChannel[IO](IO.pure(sc.channel))
+  } yield clientRes
+
+}
+```
+
+The import part here is the use of `withServerChannel`. This is a helper method
```
BenFradet:
```suggestion
The important part here is the use of `withServerChannel`. This is a helper method
```
?

Issue #826, opened 2020-03-06
gRPC server does not shutdown after `sbt run`


cb372:
If you start a server using `sbt run` (as we often do in tutorials) and then kill it with Ctrl-C, it keeps running. If you try to run it again, it fails to startup because the port is in use.

When you shutdown sbt you get an exception:

```
[error] (run-main-0) java.lang.InterruptedException
[error] java.lang.InterruptedException
[error]         at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1040)
[error]         at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1345)
[error]         at cats.effect.internals.IOPlatform$.$anonfun$unsafeResync$2(IOPlatform.scala:51)
[error]         at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
[error]         at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:57)
[error]         at scala.concurrent.package$.blocking(package.scala:146)
[error]         at cats.effect.internals.IOPlatform$.unsafeResync(IOPlatform.scala:51)
[error]         at cats.effect.IO.unsafeRunTimed(IO.scala:325)
[error]         at cats.effect.IO.unsafeRunSync(IO.scala:240)
[error]         at cats.effect.internals.IOAppPlatform$.main(IOAppPlatform.scala:25)
[error]         at cats.effect.IOApp.main(IOApp.scala:67)
[error]         at cats.effect.IOApp.main$(IOApp.scala:66)
[error]         at com.example.Server$.main(Server.scala:7)
[error]         at com.example.Server.main(Server.scala)
[error]         at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[error]         at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[error]         at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[error]         at java.base/java.lang.reflect.Method.invoke(Method.java:567)
```

cb372:
Confirmed that this is NOT fixed by #830. It's still happening.

[...]

PR #830, opened 2020-03-10
Add methods with Resource to GrpcServer and ManagedChannelInterpreter

peterneyens:
## What this does?

Add a method to `GrpcServer` and `ManagedChannelInterpreter` returning a `Resource` for shutting down the server/channel.

## Checklist


cb372:
I haven't checked, but I think this might fix #826 🤞

PR #839, opened 2020-03-11
muSrcGenIdlType Tweak

juanpedromoreno:
## What this does?

Currently, the `sbt-mu-srcgen` requires the sbt setting `muSrcGenIdlType` with a value different from the default one: `Unknown`. This happens even for those sbt projects or submodules that have no protocols to generate sources from.

```
Unknown IDL type 'Unknown'. Valid values: Proto, Avro, OpenAPI
```

As a workaround, the user can set up any of the valid values but it's not quite user-friendly.

In this PR, I'm removing the existing validation and changing it by simple check if-else. If the module has a "valid" `muSrcGenIdlType`, the plugin will run the code generator, and, in any other case, the plugin will return an empty list and, an informative message for the user.

Additionally, I'm fixing a couple of broken links in the docs.

## Checklist

### DIFF
--- a/docs/src/main/docs/guides/index.md
+++ b/docs/src/main/docs/guides/index.md
@@ -10,6 +10,6 @@ permalink: guides
 These guides are aimed at developers who are already familiar with Mu-Scala.
 
 If you are new to Mu-Scala, we recommend you read the [Getting Started
-guide](../getting-started) and the [tutorials](../tutorials).
+guide](getting-started) and the [tutorials](tutorials).
 
 Each guide introduces a single feature of Mu-Scala and explains how to use it.
--- a/docs/src/main/docs/tutorials/index.md
+++ b/docs/src/main/docs/tutorials/index.md
@@ -13,8 +13,8 @@ We recommend you read them in roughly the order they appear in the navigation
 menu.
 
 Before starting the tutorials, we recommend you read the [Getting Started
-guide](../getting-started).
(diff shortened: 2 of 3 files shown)


docs/src/main/docs/guides/index.md
```diff
@@ -10,6 +10,6 @@ permalink: guides
 These guides are aimed at developers who are already familiar with Mu-Scala.
 
 If you are new to Mu-Scala, we recommend you read the [Getting Started
-guide](../getting-started) and the [tutorials](../tutorials).
+guide](getting-started) and the [tutorials](tutorials).
```
juanpedromoreno: As the user is in the index.html, the path is omitted by the browser. It seems that it's not needed to go back with `..` .

docs/src/main/docs/tutorials/index.md
```diff
@@ -13,8 +13,8 @@ We recommend you read them in roughly the order they appear in the navigation
 menu.
 
 Before starting the tutorials, we recommend you read the [Getting Started
-guide](../getting-started).
+guide](getting-started).
 
 If you are already familiar with Mu-Scala, you may want to read about how to use
-its more advanced features in the [How-To Guides](../guides).
+its more advanced features in the [How-To Guides](guides).
```
juanpedromoreno: Same happens in this file (index.md).

PR #843, opened 2020-03-12
Fix a few warnings

### LINKED ISSUE #827: Fix compiler warnings
I went through and fixed nearly all the compiler warnings recently, but suddenly a whole load more have appeared. It's like battling the Hydra!

cb372:
Fix most of the compiler warnings.

Fix #827


modules/internal/fs2/src/main/scala/higherkindness/mu/server/fs2Calls.scala
```diff
@@ -25,6 +25,9 @@ import org.lyranthe.fs2_grpc.java_runtime.server.Fs2ServerCallHandler
 
 object fs2Calls {
 
+  // TODO Support compression. It wasn't supported in fs2-grpc
+  // at the time this code was added (#477), but it is now.
```
cb372: I'll tackle this in a separate PR

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
```diff
@@ -285,6 +285,34 @@ object serviceImpl {
...
+       * But if you write the same thing in a quasiquote inside a macro, the
+       * attachment does not get added, so you get false-positive compiler
+       * warnings at the macro use site like: "parameter value x$2 in anonymous
+       * function is never used".
+       *
+       * (The parameter needs a name, even though the function doesn't
+       * reference it, so `_` gets turned into a fresh name e.g. `x$2`.  The
+       * same thing happens even if you're not in a macro.)
+       *
+       * I'd say this is a bug in Scala. We work around it by manually adding
```
cb372: I'm going to see if I can fix this easily in scala/scala.

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
```diff
@@ -565,8 +593,9 @@ object serviceImpl {
       }
 
       val operations: List[HttpOperation] = for {
-        d      <- rpcDefs.collect { case x if findAnnotation(x.mods, "http").isDefined => x }
-        args   <- findAnnotation(d.mods, "http").collect({ case Apply(_, args) => args }).toList
+        d <- rpcDefs.collect { case x if findAnnotation(x.mods, "http").isDefined => x }
+        // TODO not sure what the following line is doing, as the result is not used. Is it needed?
+        _      <- findAnnotation(d.mods, "http").collect({ case Apply(_, args) => args }).toList
```
BenFradet: might be worth it to log an issue about it so we don't forget about it as easily, wdyt?

cb372:
@BenFradet For Scala 2.12, this PR gets us close to zero warnings, but not quite all the way there. We can't turn on `-Xfatal-warnings` just yet.

For Scala 2.13 there are lots of warnings about deprecation of `JavaConverters`, so we won't be able to turn on `-Xfatal-warnings` until we do something about that.

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
cb372: This resulted in two PRs:

* https://github.com/scala/scala/pull/8799
* https://github.com/scala/scala/pull/8803

PR #847, opened 2020-03-13
Support compression on server side with FS2 streaming

cb372:
* Make use of the compression option in `fs2Calls` (it was ignored until now)
* Refactor to pass the compression type as a proper ADT type instead of a string
* Slight tidy-up of the service macro implementation
* Enable compression on client side for compression tests
* Stop using `ConcurrentMonad` type alias in tests because it caused really long lines that messed up the formatting


cb372:
>should there be a mention of this change in the docs?

The docs should probably have mentioned that this feature wasn't implemented. But they didn't, so there is nothing to update 😄

PR #856, opened 2020-03-27
Haskell integration tests

cb372:
Add a suite of integration tests to check RPC communication between Mu-Scala and Mu-Haskell.

See modules/haskell-integration-tests/README.md for more details about how the tests work.

These tests have already proved quite useful. While writing them I discovered a few issues on the Haskell side:

* https://github.com/higherkindness/mu-haskell/pull/144
* https://github.com/higherkindness/mu-haskell/pull/156
* https://github.com/haskell-grpc-native/http2-grpc-haskell/pull/20
* https://github.com/higherkindness/mu-haskell/issues/157
* https://github.com/higherkindness/mu-haskell/issues/163

I also learned a lot about gRPC compression. I have opened a PR (https://github.com/higherkindness/mu-scala/pull/857) to improve our docs in this area.

For now there are only tests for Protobuf, and half of them are marked as pending because of an issue on the Haskell side. I intend to add tests for streaming endpoints and Avro later.

Regarding sbt/Travis, I decided not to include the new module in the aggregate, and to run the tests as a separate Travis job so they can run in parallel. But I'm happy to discuss alternative configurations.


modules/haskell-integration-tests/mu-haskell-client-server/protobuf-client/Main.hs
```diff
@@ -0,0 +1,52 @@
...
+      d = readMaybe days
+      req = record (c, d)
+
+showGetForecastResponse :: GetForecastResponse -> String
+showGetForecastResponse resp =
+  lastUpdated ++ " " ++ (intercalate ", " dailyForecasts)
+    where
+      lastUpdated = unpack(fromMaybe "" (resp ^. #last_updated))
+      -- TODO actually I want to print just "SUNNY" here, but it's too hard
+      dailyForecasts = show <$> (fromMaybe [] (resp ^. #daily_forecasts))
```
cb372: @serras Could you help with this? How can I get just the enum choice name as a string? I had a look at the [Show instance](https://hackage.haskell.org/package/mu-schema-0.2.0.0/docs/src/Mu.Schema.Interpretation.html#line-191). While I understand what it's doing, my type-level-haskell-fu was not strong enough to adapt it to my needs.

Actually I've just noticed there are `toJSON` instances available. It might make sense for the client to print the response to stdout as JSON. That would be easy for the tests to parse.

kutyel:
Awesome work Chris! 🚀 🙌 🔥

modules/haskell-integration-tests/mu-haskell-client-server/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM fpco/stack-build:lts-14.27 as build
```
kutyel: Maybe you can use here the latest lts we just migrated to!
```suggestion
FROM fpco/stack-build:lts-15.5 as build
```

modules/haskell-integration-tests/mu-haskell-client-server/Dockerfile.warm_dot_stack
```diff
@@ -0,0 +1,10 @@
+# This Dockerfile runs 'stack build', thus creating an image
+# containing a warm ~/.stack directory.
+# This is used to speed up the building of the actual image
+# we want to use for the tests.
+
+FROM fpco/stack-build:lts-14.27
```
kutyel: Here too

modules/haskell-integration-tests/mu-haskell-client-server/Dockerfile
cb372: The tests are currently using mu-haskell 0.2, which uses `lts-14.27`. Maybe it makes sense to upgrade the LTS at the same time we upgrade the tests to mu-haskell 0.3?

modules/haskell-integration-tests/mu-haskell-client-server/protobuf-client/Main.hs
cb372: Never mind, I used Aeson to render the list of enum values as JSON.

modules/haskell-integration-tests/mu-haskell-client-server/Dockerfile
kutyel: That makes sense! 😄

PR #866, opened 2020-04-08
Distributed tracing with Natchez

cb372:
This PR integrates [Natchez](https://github.com/tpolecat/natchez) to provide support for distributed tracing of Mu-Scala applications.

## Client side

Currently the standard way to create a client for a Mu service `MyService[F]` is the macro-generated `MyService.client` function:

```scala
def client[F[_]: ConcurrentEffect: ContextShift](
  channelFor: ChannelFor,
  channelConfigList: List[ManagedChannelConfig] = List(UsePlaintext()),
  options: CallOptions = CallOptions.DEFAULT
): Resource[F, MyService[F]]
```

The returned `Resource` takes care of creating and tearing down the channel (i.e. the underlying HTTP client), and when you `use` it you get access to a client that returns responses in the effect monad `F`, e.g. cats-effect `IO`.

This PR adds a similar method for creating a client with Natchez tracing support:

```scala
def tracingClient[F[_]: ConcurrentEffect: ContextShift](
  channelFor: ChannelFor,
  channelConfigList: List[ManagedChannelConfig] = List(UsePlaintext()),
  options: CallOptions = CallOptions.DEFAULT
): Resource[F, MyService[Kleisli[F, Span[F], *]]]
```

Again it returns a `Resource[F, ...]`, which takes care of setting up the channel. But this time `use`-ing the `Resource` gives you a `MyService[Kleisli[F, Span[F], *]]`. All the methods of the client have a return type of the form `Kleisli[F, Span[F], Response]`, meaning "if you give me the current span, I'll return you an `F[Response]`".

As a concrete example, let's use good old `SayHello`:

```scala
trait MyService[F[_]] {
  def SayHello(req: HelloRequest): F[HelloResponse]
}
```

and assume we are using cats-effect `IO`:

```scala
val clientResource: Resource[IO, MyService[Kleisli[IO, Span[IO], *]]] =
  MyService.tracingClient[IO]

resource.use { client =>
  val currentSpan: Span[IO] = ???
  val request = HelloRequest()
  val kleisli: Kleisli[IO, Span[IO], HelloResponse] = client.SayHello(request)
  val response: IO[HelloResponse] = kleisli.run(currentSpan)
  // ...
}
```

See the example code (linked below) for details of how to get hold of the current span.

Given the current span, the client does two things:

1. Create a child span, with the fully qualified name of the RPC endpoint it's calling
2. Add the tracing-related headers to the RPC request, so the server can continue the trace

## Server side

Currently the standard way to create a server for a Mu service `MyService[F]` is to write an interpreter of the `MyService[F]` algebra and then bind it using the the macro-generated `MyService.bindService` function:

```scala
def bindService[F[_]: ConcurrentEffect](implicit
  algebra: MyService[F]
): F[ServerServiceDefinition]
```

This PR adds a similar function for binding a tracing service:

```scala
def bindTracingService[F[_]: ConcurrentEffect](
  entrypoint: EntryPoint[F]
)(implicit 
  algebra: MyService[Kleisli[F, Span[F], *]]
): F[ServerServiceDefinition]
```

As an explicit argument it takes a `natchez.EntryPoint[F]`, which is used to create root spans (i.e. new traces) or to continue traces based on incoming headers.

As an implicit argument it takes an algebra of type `MyService[Kleisli[F, Span[F], *]]`.

This type looks a bit scary, but within user code there's no need to worry about `Kleisli`. The user can write their code in terms of `F[_]: Monad` or whatever, just like they normally would. If they want to create child spans, they can use Natchez's `Trace` type class.

For example a user-written implementation of `MyService` might look like:

```scala
class MyTracingService[F: Trace: Applicative] extends MyService[F] {

  def SayHello[F](req: HelloRequest): F[HelloResponse] =
    Trace[F].span("here's a child span") {
      HelloResponse("hi there!").pure[F]
    }

}
```

The macro-generated server call handlers do 2 things:

1. Extract any tracing-related headers from the RPC request
2. Create a span with the fully-qualified name of the endpoint being called
    a. If tracing-related headers were found, it continues the trace that was started upstream. 
    b. Otherwise it starts a new trace, using the endpoint name to create the root span

## Example

Here is a trace across 3 applications.

Screenshot 2020-04-08 at 10 47 49

The working example code that generated this trace is available here: https://github.com/cb372/mu-tracing-example. See the README in that repo for details on how to run it.

Once this feature is merged and released, we could add the example app to https://github.com/higherkindness/mu-scala-examples.

## Problem: response streaming

Consider a service with a variety of streaming endpoints:

```scala
trait StreamingService[F[_]] {
  def unary(req: Request): F[Response]
  def clientStreaming(reqs: Stream[F, Request]): F[Response]
  def serverStreaming(req: Request): Stream[F, Response]
  def bidirectional(reqs: Stream[F, Request]): Stream[F, Response]
}
```

(In the case of Monix, replace all `Stream[F, A]` with `Observable[A]`).

The first two endpoints work fine with tracing. The return types have the shape `F[*]`, so in the tracing client the return type becomes a `Kleisli` (e.g. `Kleisli[IO, Span[IO], Response]`). That gives us the access to the current span that we need in order to add the tracing headers to the request.

However, the endpoints that return `Stream[F, Response]` (or for Monix, `Observable[Response]`) do not have the required shape. There is no way to turn that return type into a Kleisli.

For now I've left stub implementations that just throw exceptions if you try to use tracing with server-streaming or bidirectional-streaming endpoints.

I recommend we change Mu's service spec slightly to ensure all return types are of the shape `F[*]`. So the service above would need to be rewritten to:

```scala
trait StreamingService[F[_]] {
  def unary(req: Request): F[Response]
  def clientStreaming(reqs: Stream[F, Request]): F[Response]
  def serverStreaming(req: Request): F[Stream[F, Response]]
  def bidirectional(reqs: Stream[F, Request]): F[Stream[F, Response]]
}
```

This would require the following work for us:
1. Update the code generation in skeuomorph
2. Update the macro to expect return types with the new shape

And it would be a breaking change in Mu, resulting in some trivial migration work for users. In their server-side implementation of `StreamingService` they would just need to wrap their response streams in `pure(...)` to lift them into `F`.

Does this sound reasonable? If so, I'll put this PR on hold and get started on making that change.

## Implementation notes

Originally I expected to add a new `mu-tracing` module, which people would add to their build if they wanted to make use of tracing. I was hoping we could just add some interceptors, similar to the way the metrics module works. But this was not really feasible. It doesn't fit the way Natchez works, using `Kleisli`, and it doesn't give us the read/write access to the request headers that we need.

I ended up adding the tracing support into the `internal-core` module. So that module now has a dependency on `natchez-core`. (Users need to add the appropriate natchez module for their chosen tracing implementation, e.g. `natchez-jaeger`.) Adding such an immature library (currently version 0.0.11) as a core dependency is not ideal. But it's a tiny library, so there's not that much to break, and any bincompat-breaking changes should only affect users who have chosen to use tracing, not all Mu users.

## TODO

- [x] Agree on the approach for response-streaming endpoints
- [x] Implement tracing versions of FS2 response-streaming and bidirectional-streaming endpoints
- [x] Implement tracing versions of Monix streaming endpoints
- [x] Implement `tracingClientFromChannel`, `unsafeTracingClient`, `unsafeTracingClientFromChannel` in the macro
- [x] Write tests
- [ ] Add a how-to guide to the docs
- [ ] (after next release) Add the example to the examples repo


modules/internal/src/main/scala/higherkindness/mu/rpc/internal/server/TracingUnaryServerCallHandler.scala
```diff
@@ -0,0 +1,59 @@
...
+  ): Listener[Req] = {
+    val asciiHeaders = metadata
+      .keys()
+      .asScala
+      .collect {
+        case k if !k.endsWith(Metadata.BINARY_HEADER_SUFFIX) =>
+          k -> metadata.get(Key.of(k, Metadata.ASCII_STRING_MARSHALLER))
+      }
+      .toMap
+    val kernel       = Kernel(asciiHeaders)
```
BenFradet, line 48: this seems to be the same as `extractTracingKernel` below

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/server/package.scala
```diff
@@ -56,4 +59,16 @@ package object server {
       }
   }
 
+  private[internal] def extractTracingKernel(headers: Metadata): Kernel = {
+    val asciiHeaders = headers
+      .keys()
+      .asScala
+      .collect {
+        case k if !k.endsWith(BINARY_HEADER_SUFFIX) =>
```
BenFradet, line 67: dumb question, is it that all internal headers are, if any, binary?

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/server/TracingUnaryServerCallHandler.scala
cb372: Well spotted, that's a refactoring mistake.

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/server/package.scala
cb372: gRPC specifies that all custom headers must include the `-bin` suffix in their names if the header value is binary, and NOT include the suffix if the value is valid ASCII. So a server knows, just by looking at the header name, how to deserialize the value.

I don't think it's related to whether they are "internal" headers or not.

All the gory details are in the spec: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests

BenFradet: Thanks 👍

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/client/HeaderAttachingClientCall.scala
```diff
@@ -0,0 +1,33 @@
...
+import io.grpc.ForwardingClientCall.SimpleForwardingClientCall
+
+class HeaderAttachingClientCall[Req, Res](
+    call: ClientCall[Req, Res],
+    extraHeaders: Metadata
+) extends SimpleForwardingClientCall[Req, Res](call) {
+
+  override def start(responseListener: Listener[Res], headers: Metadata): Unit = {
+    headers.merge(extraHeaders)
+    super.start(responseListener, headers)
```
fedefernandez, line 30: :+1:

tzimisce012:
Great job 👍

cb372:
The implementation and tests are now complete, so this PR is ready for another round of review. The only thing left to do is write the how-to guide.

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
```diff
@@ -255,6 +255,13 @@ object serviceImpl {
         s"@service-annotated class $serviceName's type parameter must be higher-kinded"
       )
 
+      // Type lambda for Kleisli[F, Span[F], *]
+      private val kleisliFSpanF =
+        tq"({ type T[α] = _root_.cats.data.Kleisli[$F, _root_.natchez.Span[$F], α] })#T"
```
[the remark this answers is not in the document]
cb372: Yes, I don't think we should make kind-projector mandatory for end-users. And if we relied on kind-projector but the user forgot to enable it, they would probably get a very confusing compiler error.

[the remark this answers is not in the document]
BenFradet: ok, my concern was just around making it more readable.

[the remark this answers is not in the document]
BenFradet: but obviously since they should have to interact with it much, it makes sense to not make it mandatory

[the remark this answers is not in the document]
cb372: Actually I just realised we definitely can't use kind-projector syntax in macros. The kind-projector plugin runs between the `parser` and `namer` phases (i.e. before typechecking), but the macro is expanded during typechecking.

So if we generated a tree corresponding to the kind-projector syntax `Kleisli[F, Span[F], *]`, the kind-projector plugin would not have a chance to rewrite it into a type-projection tree.

modules/internal/fs2/src/main/scala/higherkindness/mu/rpc/internal/server/fs2/handlers.scala
```diff
@@ -0,0 +1,141 @@
...
+      f: (Stream[F, Req], Metadata) => F[Stream[F, Res]],
+      compressionType: CompressionType
+  ): ServerCallHandler[Req, Res] =
+    Fs2ServerCallHandler[F].streamingToStreamingCall[Req, Res](
+      (stream, metadata) => Stream.eval(f(stream, metadata)).flatten,
+      serverCallOptions(compressionType)
+    )
+
+  private type Traced[F[_], A]         = Kleisli[F, Span[F], A]
+  private type StreamOfTraced[F[_], A] = Stream[Kleisli[F, Span[F], *], A]
```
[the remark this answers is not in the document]
cb372: I tried that initially but I decided to remove the aliases and use raw Kleisli as much as possible. When I had aliases I found myself having to mentally un-alias whenever I read the types, so it added cognitive overhead instead of reducing it.

I only kept these aliases to make the type signatures shorter in this file because they were sending the formatter crazy.

PR #871, opened 2020-04-14
Wrap streaming responses in effect

cb372:
This is a change to the specification for Mu services. Previously, RPC endpoints with streaming responses were encoded as:

```scala
// with fs2
def foo(req: MyRequest): Stream[F, MyResponse]
// with Monix
def foo(req: MyRequest): Observable[MyResponse]
```

Now we require them to be inside `F[_]`, making them consistent with endpoints that return unary responses. They now look like:

```scala
// with fs2
def foo(req: MyRequest): F[Stream[F, MyResponse]]
// with Monix
def foo(req: MyRequest): F[Observable[MyResponse]]
```

This change is partly for consistency's sake, and mostly because it's required for distributed tracing (see #866 for details).

It's a breaking change for any users who use streaming responses, but the required fix in user code should be pretty small. On the server side, their service implementation just needs to lift the result into `F[_]` using `pure`. For example:

```scala
class MyServiceImpl[F[_]] extends MyService[F] {
  def foo(req: MyRequest): Stream[F, MyResponse] =
    Stream(MyResponse("hello"), MyResponse("world"))
}
```

becomes:

```scala
import cats.syntax.applicative._

class MyServiceImpl[F[_]: Applicative] extends MyService[F] {
  def foo(req: MyRequest): Stream[F, MyResponse] =
    Stream(MyResponse("hello"), MyResponse("world")).pure[F]
}
```

On the client side, they need to deal with the fact that the client method now returns `F[Stream[F, MyResponse]]` instead of a plain `Stream[F, MyResponse]`. With FS2, the easiest way to handle this is to use `Stream.eval` to make a stream-of-streams, then `flatten` it:

```scala
val res: F[Stream[F, MyResponse]] = client.foo(req)
val streamOfStreams: Stream[F, Stream[F, MyResponse]] = Stream.eval(res)
val stream: Stream[F, MyResponse] = streamOfStreams.flatten
```

With Monix, the corresponding incantation is:

```scala
val res: F[Observable[MyResponse]] = client.foo(req)
val nestedObs: Observable[Observable[MyResponse]] = Observable.from(res)
val obs: Observable[MyResponse] = obs.flatten
```

Potentially we could publish scalafix rewrites to help users migrate, but I'm not sure it's worth the effort.

## Other changes

As part of this work I had to refactor the macro code slightly. I also added a few more compile-time checks to the macro, and I ended up improving the support for `Empty` requests and responses in the HTTP-related macro code.

### DIFF
--- a/modules/channel/src/test/scala/higherkindness/mu/rpc/MacroAnnotationCompilationTest.scala
+++ b/modules/channel/src/test/scala/higherkindness/mu/rpc/MacroAnnotationCompilationTest.scala
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2017-2020 47 Degrees, LLC. <http://www.47deg.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package higherkindness.mu.rpc
+
+import higherkindness.mu.rpc.protocol._
+import fs2.Stream
+import monix.reactive.Observable
+
+// This is not a normal "test" with assertions.
+// It's just a check that valid code compiles as expected
+// after the @service macro has been expanded.
+object MacroAnnotationCompilationTest {
+
+  final case class MyReq()
+  final case class MyResp(a: Int)
+
+  @service(Protobuf)
+  trait MyFS2Service[F[_]] {
+
+    def trivial(req: Empty.type): F[Empty.type]
+
+    def unary(req: MyReq): F[MyResp]
+
+    def clientStreaming(req: Stream[F, MyReq]): F[MyResp]
+
+    def clientStreamingFullyQualifiedStream(req: _root_.fs2.Stream[F, MyReq]): F[MyResp]
+
+    def serverStreaming(req: MyReq): F[Stream[F, MyResp]]
+
+    def bidirectionalStreaming(req: Stream[F, MyReq]): F[Stream[F, MyResp]]
+
+  }
+
+  @service(Protobuf)
+  trait MyMonixService[F[_]] {
+
+    def trivial(req: Empty.type): F[Empty.type]
+
+    def unary(req: MyReq): F[MyResp]
+
+    def clientStreaming(req: Observable[MyReq]): F[MyResp]
+
+    def clientStreamingFullyQualifiedStream(req: Observable[MyReq]): F[MyResp]
+
+    def serverStreaming(req: MyReq): F[Observable[MyResp]]
+
+    def bidirectionalStreaming(req: Observable[MyReq]): F[Observable[MyResp]]
+
+  }
+
+}
--- a/modules/http/src/test/scala/higherkindness/mu/rpc/http/GreeterDerivedRestTests.scala
+++ b/modules/http/src/test/scala/higherkindness/mu/rpc/http/GreeterDerivedRestTests.scala
@@ -116,9 +116,9 @@ class GreeterDerivedRestTests extends RpcBaseTestSuite with BeforeAndAfter {
     "serve a POST request with fs2 streaming response" in {
       val request = HelloRequest("hey")
       val responses =
-        BlazeClientBuilder[IO](ec).stream.flatMap(client =>
-          Stream.eval(fs2Client.sayHelloAll(request)(client)).flatten
-        )
+        BlazeClientBuilder[IO](ec).stream
+          .evalMap(client => fs2Client.sayHelloAll(request)(client))
+          .flatten
       responses.compile.toList
         .unsafeRunSync() shouldBe List(HelloResponse("hey"), HelloResponse("hey"))
     }
--- a/modules/http/src/test/scala/higherkindness/mu/rpc/http/GreeterDerivedRestTests.scala
+++ b/modules/http/src/test/scala/higherkindness/mu/rpc/http/GreeterDerivedRestTests.scala
@@ -53,8 +53,8 @@ class GreeterDerivedRestTests extends RpcBaseTestSuite with BeforeAndAfter {
 
   "REST Service" should {
 
-    val unaryClient = UnaryGreeter.httpClient[IO](serviceUri)
-    val fs2Client   = Fs2Greeter.httpClient[IO](serviceUri)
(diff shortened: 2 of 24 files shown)


modules/channel/src/test/scala/higherkindness/mu/rpc/MacroAnnotationCompilationTest.scala
```diff
@@ -0,0 +1,65 @@
...
+  @service(Protobuf)
+  trait MyFS2Service[F[_]] {
+
+    def trivial(req: Empty.type): F[Empty.type]
+
+    def unary(req: MyReq): F[MyResp]
+
+    def clientStreaming(req: Stream[F, MyReq]): F[MyResp]
+
+    def clientStreamingFullyQualifiedStream(req: _root_.fs2.Stream[F, MyReq]): F[MyResp]
```
juanpedromoreno, line 40: Just in case it wasn't intentional:

```suggestion
    def clientStreamingFullyQualifiedStream(req: Stream[F, MyReq]): F[MyResp]
```

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
```diff
@@ -28,15 +28,28 @@ object serviceImpl {
...
+     *
+     * @unwrappedType The type, with any surrounding effect type stripped.
+     *                e.g. `MyRequest` or `Stream[F, MyRequest]` or `Observable[MyRequest]`
+     *                For a request type, `unwrappedType` == `originalType`.
+     *
+     * @messageType The type of the message in a request/response.
+     *              For non-streaming request/responses, `messageType` == `unwrappedType`.
+     *              For streaming request/responses, `messageType` is the type of the stream elements.
+     *              e.g. if `originalType` is `F[Stream[F, MyResponse]]` then `messageType` is `MyResponse`.
+     */
```
juanpedromoreno, line 46: Nicely documented 👍

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
```diff
@@ -50,19 +63,55 @@ object serviceImpl {
...
+                )
+            }
+          } else {
+            // Request type is not wrapped in F[...], so return it as-is
+            t
+          }
+        }
+
+        unwrappedType match {
+          case tq"$tpe[$elemType]" if identOrSelect(tpe, "Observable") =>
```
juanpedromoreno, line 91: Would we need FQN here and for the `Stream` line below?

modules/http/src/test/scala/higherkindness/mu/rpc/http/GreeterDerivedRestTests.scala
```diff
@@ -116,31 +116,39 @@ class GreeterDerivedRestTests extends RpcBaseTestSuite with BeforeAndAfter {
...
         .unsafeRunSync() should have message "java.lang.IllegalArgumentException: empty greeting"
     }
 
     "serve a POST request with bidirectional fs2 streaming" in {
       val requests = Stream(HelloRequest("hey"), HelloRequest("there"))
       val responses =
-        BlazeClientBuilder[IO](ec).stream.flatMap(fs2Client.sayHellosAll(requests)(_))
+        BlazeClientBuilder[IO](ec).stream.flatMap(client =>
+          Stream.eval(fs2Client.sayHellosAll(requests)(client)).flatten
+        )
```
BenFradet, line 141: can't this be:

```scala
BlazeClientBuilder[IO](ec).stream,evalMap { client =>
  fs2Client.sayHellosAll(requests)(client)
}.flatten
```

modules/http/src/test/scala/higherkindness/mu/rpc/http/GreeterHandlers.scala
```diff
@@ -50,10 +52,22 @@ class Fs2GreeterHandler[F[_]: Sync] extends Fs2Greeter[F] {
...
+  def sayHelloAll(request: HelloRequest): F[Stream[F, HelloResponse]] =
+    Applicative[F].pure(
+      if (request.hello.isEmpty) Stream.raiseError(new IllegalArgumentException("empty greeting"))
+      else Stream(HelloResponse(request.hello), HelloResponse(request.hello))
+    )
+
+  def sayHelloAllEmptyRequest(request: Empty.type): F[Stream[F, HelloResponse]] =
+    Applicative[F].pure(
+      Stream(HelloResponse("hey"), HelloResponse("hi again"))
+    )
```
BenFradet, line 67: wouldn't the following be a bit more readable?

```scala
Stream(HelloResponse("hey"), HelloResponse("hi again")).pure[F]
```

modules/channel/src/test/scala/higherkindness/mu/rpc/MacroAnnotationCompilationTest.scala
cb372: This was intentional. It's testing the behaviour of the `identOrSelect` function.

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
cb372: `identOrSelect` will match `Observable`, both qualified and unqualified. So it will match any of:

* `_root_.monix.reactive.Observable`
* `monix.reactive.Observable`
* `reactive.Observable`
* `Observable`

Unfortunately it will also match `com.mylibrary.Observable`, which is not ideal.

I'll add a comment to the code to explain what it's doing, and try rewriting it to be stricter.

Unfortunately we don't have access to any types at this point, so we have to do the best we can with only syntactic analysis.

modules/http/src/test/scala/higherkindness/mu/rpc/http/GreeterDerivedRestTests.scala
cb372: Yep, that's a bit cleaner 👍

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
cb372: Rewritten and commented in 768a3e476bd6cf1182aee6cdb6cb0cb0b21b4564

modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala
```diff
@@ -64,11 +64,44 @@ object serviceImpl {
...
+       *
+       * {{{
+       * val fqn = List("_root_", "monix", "reactive", "Observable")
+       *
+       * possiblyQualifiedName(tq"Observable", fqn) == true
+       * possiblyQualifiedName(tq"reactive.Observable", fqn) == true
+       * possiblyQualifiedName(tq"monix.reactive.Observable", fqn) == true
+       * possiblyQualifiedName(tq"_root_.monix.reactive.Observable", fqn) == true
+       * possiblyQualifiedName(tq"com.mylibrary.Observable", fqn) == false
+       * }}}
```
juanpedromoreno, line 95: 👍

PR #884, opened 2020-04-21
Using a "real" gRPC Server for Benchmarks

juanpedromoreno:
## What this do?

This PR changes the _in-memory_ server for a real Netty server for running the benchmarks. They have become slower, but they are now more accurate.

Additionally, to try to speed them up a bit, I've decreased the number of iterations (and warm-up iterations) from 10 to 5.

Fixes https://github.com/higherkindness/mu-scala/issues/870, since the performance restores in this way.
Related to https://github.com/grpc/grpc-java/issues/6926.

## Checklist

- [x] Reviewed the diff to look for typos, println, and format errors.


juanpedromoreno:
Currently, the build is failing with:

```bash
Installing Java version 'adopt@1.8'
Install Java
  exec curl -sL -o /home/runner/bin/jabba https://github.com/shyiko/jabba/releases/download/0.11.2/jabba-0.11.2-linux-amd64 { silent: *** }
  chmod 755 /home/runner/bin/jabba
  exec /home/runner/bin/jabba ls-remote
  /home/runner/bin/jabba: 2: /home/runner/bin/jabba: Syntax error: newline unexpected
  ##[error]exec: /home/runner/bin/jabba: 2: /home/runner/bin/jabba: Syntax error: newline unexpected
```

I'll check it again tomorrow once the Github service is back to normal.

benchmarks/shared/src/main/scala/higherkindness/mu/rpc/benchmarks/AvroBenchmark.scala
```diff
@@ -18,52 +18,52 @@ package higherkindness.mu.rpc.benchmarks
...
-      .flatMap(sc => PersonServiceAvro.clientFromChannel[IO](IO(sc.channel)))
+  def clientCall[B](f: PersonServiceAvro[IO] => IO[B]): B =
+    PersonServiceAvro
+      .client[IO](ChannelForAddress("localhost", grpcPort))
+      .use(f)
+      .unsafeRunTimed(defaultTimeOut)
+      .get
 
   @TearDown
   def shutdown(): Unit = {}
```
cb372, line 46: Shouldn't we shut the server down here, as it's now a "real" server bound to a port?

juanpedromoreno: Fixed in 964cde1

benchmarks/shared/src/main/scala/higherkindness/mu/rpc/benchmarks/AvroBenchmark.scala
```diff
@@ -18,52 +18,52 @@ package higherkindness.mu.rpc.benchmarks
...
+    PersonServiceAvro
+      .client[IO](ChannelForAddress("localhost", grpcPort))
+      .use(f)
+      .unsafeRunTimed(defaultTimeOut)
+      .get
 
   @Benchmark
   def listPersons: PersonList =
-    clientIO.use(_.listPersons(Empty)).unsafeRunTimed(defaultTimeOut).get
+    clientCall(_.listPersons(Empty))
```
cb372, line 50: This builds (and tears down) the client every time it's called, which involves creating object pools and various other expensive things. I didn't look too closely but there's a lot going on in [AbstractManagedChannelImplBuilder](https://github.com/grpc/grpc-java/blob/459cbc7b44e8e14748dbfbd4c47a2958f34778e9/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java), so it's probably pretty slow.

Maybe a more realistic benchmark would be to create the client once during the setup phase and re-use it. That matches what most users would do in production: create the `Resource` and call `use` on it once at the edge of their app, and pass the client around wherever it's needed.

Shall we take a look at doing this in a separate PR?

benchmarks/shared/src/main/scala/higherkindness/mu/rpc/benchmarks/shared/ServerRuntime.scala
```diff
@@ -0,0 +1,70 @@
...
+      _ <- logger.info("Starting server..")
+      _ <- Resource.liftF(server.start()).use(_ => IO.never).start
+      _ <- logger.info("Server started..")
+    } yield ()).unsafeRunSync
+
+  def stopServer(implicit server: GrpcServer[IO]): Unit =
+    (for {
+      _ <- logger.info("Stopping server..")
+      _ <- server.shutdownNow()
+      _ <- logger.info("Server Stopped..")
```
cb372, line 68: You might want to add a `server.awaitTermination()` here just in case

juanpedromoreno: Done in d5248bc

benchmarks/shared/src/main/scala/higherkindness/mu/rpc/benchmarks/AvroBenchmark.scala
juanpedromoreno: Thanks Chris, you are completely right.

I've tried to fix it in https://github.com/higherkindness/mu-scala/pull/884/commits/7957b3ff85e01b2d981610d9dc71c44162a94a67, what do you think about doing it that way?

benchmarks/shared/src/main/scala/higherkindness/mu/rpc/benchmarks/shared/ServerRuntime.scala
```diff
@@ -0,0 +1,86 @@
...
+      List(UsePlaintext())
+    ).build.unsafeRunSync
+
+  implicit lazy val grpcServer: GrpcServer[IO] =
+    grpcConfigs.flatMap(conf => GrpcServer.default[IO](grpcPort, conf)).unsafeRunSync
+
+  def startServer(implicit server: GrpcServer[IO]): Unit =
+    (for {
+      _ <- logger.info("Starting server..")
+      _ <- Resource.liftF(server.start()).use(_ => IO.never).start
```
cb372, line 71: Can't this line be simplified to just `_ <- server.start()` ?

benchmarks/shared/src/main/scala/higherkindness/mu/rpc/benchmarks/AvroBenchmark.scala
cb372: LGTM 👍 I'm interested to see the benchmark results after this change.

juanpedromoreno: They seem to be a bit faster now :)

benchmarks/shared/src/main/scala/higherkindness/mu/rpc/benchmarks/shared/ServerRuntime.scala
juanpedromoreno: Done in https://github.com/higherkindness/mu-scala/pull/884/commits/5e703b1259248c401a89b35b9708c1201f2d2d5b

PR #887, opened 2020-04-22
Refactor the `@service` macro

### LINKED ISSUE #844: Review the HTTP-related code in serviceImpl.scala
https://github.com/higherkindness/mu-scala/blob/725e46f4c0adeab8d87c2bab76476ed273254677/modules/internal/src/main/scala/higherkindness/mu/rpc/internal/serviceImpl.scala#L597-L598

### LINKED ISSUE #874: Refactor the macro
Once #866 is merged, `serviceImpl.scala` will be over 1100 lines long. Wild!

cb372:
Split the monolithic macro implementation into a few files.

Note: to help the compiler out with the path-dependent types arising from use of `Context`, the helper classes must be of the form

```scala
class MyHelper[C <: Context](val c: C)
```

and when you instantiate the helper, you need to give it the `Context` instance's singleton type:

```scala
val myHelper = new MyHelper[c.type](c)
```

Ideally I'd like the `Operation`, `RPCMethod` and `HttpOperation` classes to be in separate files, but I started to hit the limits of scalac's path-dependent types support so I had to group them together.

Potentially a few more utility methods could be moved to the `TreeHelpers` class, but this refactoring is a good start.

Also tidied up the HTTP-related macro code:
* removed an unnecessary line from the `val httpOperations` for-comprehension
* fixed some code that relied on the service definition trait's type parameter being called `F[_]` (the macro would have failed if the user decided to name their type param `G[_]`)

Fixes #874, fixes #844


modules/internal/src/main/scala/higherkindness/mu/rpc/internal/service/makro/OperationModels.scala
```diff
@@ -0,0 +1,477 @@
...
+      case Some(BidirectionalStreaming) =>
+        q"""
+        def $name(input: $reqType): $wrappedRespType =
+          ${clientCallMethodFor("bidiStreaming")}
+        """
+      case None =>
+        q"""
+        def $name(input: $reqType): $wrappedRespType =
+          ${clientCallMethodFor("unary")}
+        """
```
juanpedromoreno, line 159: Seems like the four options here are identical, except the `clientMethodName`, maybe we can simplify the pattern matching to get that information and leave the rest outside just once?

PR #949, opened 2020-06-19
Update Haskell code after breaking change in Mu Haskell

serras:
## What this does?

Fixes some Haskell code required by changes in upstream Mu-Haskell.

## Checklist

- [X] Reviewed the diff to look for typos, println and format errors.


modules/haskell-integration-tests/mu-haskell-client-server/protobuf-server/Main.hs
```diff
@@ -60,7 +60,7 @@ subscribeToRainEvents req sink = runConduit $ C.yieldMany events .| sink
   toRainEvent x = record (city, Just x)
   city = req ^. #city
 
-server :: (MonadServer m) => SingleServerT WeatherService m _
+server :: (MonadServer m) => SingleServerT info WeatherService m _
```
cb372, line 63: Do we need the same change in `avro-server/Main.hs`?

serras: Done!

PR #966, opened 2020-07-03
Bumps http4s

fedefernandez:
## What this does?

Bumps http4s

## Checklist


fedefernandez:
I've changed the `bodyAsText` as suggested in the docs but now the circe decoding is failing. I need to take a deep look to this

modules/http/src/main/scala/higherkindness/mu/http/implicits.scala
```diff
@@ -65,7 +65,10 @@ object implicits {
         decoder: Decoder[A],
         F: ApplicativeError[F, Throwable]
     ): Stream[F, A] =
-      message.body.chunks.parseJsonStream.map(_.as[A]).rethrow
+      message.body.chunks.parseJsonStream
+        .flatMap(
+          _.as[List[A]].map(Stream.emits(_)).fold(_.raiseError[Stream[F, *], A], identity(_))
```
fedefernandez, line 70: Need to investigate this. I know http4s changed at some point the streaming approach and started to return chunks of data instead of an array. For some reason, we're receiving here an array, instead of a single element.

project/ProjectPlugin.scala
```diff
@@ -27,7 +27,7 @@ object ProjectPlugin extends AutoPlugin {
       val fs2Kafka: String              = "1.0.0"
       val grpc: String                  = "1.30.2"
       val jodaTime: String              = "2.10.6"
-      val http4s: String                = "0.21.0-M6"
+      val http4s: String                = "0.21.6"
```
juanpedromoreno, line 30:
```suggestion
      val http4s: String                = "0.21.7"
```

juanpedromoreno (approved):

modules/http/src/main/scala/higherkindness/mu/http/implicits.scala
fedefernandez: @juanpedromoreno this isn't optimal, we should be receiving a single element in the stream, not a list. I couldn't find a gap to further investigate.

bilki, line 70: What if we use this? It's almost the same, but single elements are emitted by `jawn-fs2` instead of handling that ourselves. I didn't find exactly why chunks are now received this way, there was a change in how chunks are backed (predef `Seq` between Scala 2.12 and 2.13 changed from mutable to immutable) but nothing specific to this...

```suggestion
      message.body.chunks.unwrapJsonArray.map(_.as[A]).rethrow
```

Tests seem to be passing.

fedefernandez: 🤔  interesting. My concern is why we're receiving the array instead of separated elements. That could potentially affect to the performance unless I'm missing something, like data be grouped in chunks due to how they are generated. Anyway, your change makes sense to me.

mintrabanno:
@Banno/jabberwocky

juanpedromoreno (dismissed):

juanpedromoreno (approved):

PR #969, opened 2020-07-09
add instructions for manually enabling the sbt-mu-srcgen plugin for versions beyond v0.23.x

dmarticus:
## What this does?

With this PR (higherkindness/sbt-mu-srcgen#72), we'll need to update the microsite for mu-scala to clarify the correct usage of the sbt-mu-srcgen, since the above change makes it so the plugin is disabled by default and that the users of that plugin will need to update their `build.sbt` files accordingly to take advantage of the new functionality.  

This change is the final piece of the puzzle for this issue: https://github.com/higherkindness/sbt-mu-srcgen/issues/38

## Checklist

### DIFF
--- a/microsite/src/main/docs/reference/source-generation.md
+++ b/microsite/src/main/docs/reference/source-generation.md
@@ -14,6 +14,8 @@ Scala source code from Avro/Protobuf/OpenAPI IDL files.
 
 This section explains each of the sbt plugin's settings.
 
+TODO: should I add a section explaining how this plugin is disabled by default, or is that more appropriate in the other settings?
+
 ### muSrcGenIdlType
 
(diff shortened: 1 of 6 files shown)


microsite/src/main/docs/reference/source-generation.md
```diff
@@ -14,6 +14,8 @@ Scala source code from Avro/Protobuf/OpenAPI IDL files.
 
 This section explains each of the sbt plugin's settings.
 
+TODO: should I add a section explaining how this plugin is disabled by default, or is that more appropriate in the other settings?
```
juanpedromoreno, line 17: At this point, you can add a reminder that the sbt plugin needs to be enabled, so something like the note above would work too.

dmarticus: works for me!

PR #1042, opened 2020-11-03
Update the integration tests to use mu-haskell v0.4

cb372:
I've pushed the new images to dockerhub:

- cb372/mu-haskell-warm-dot-stack:lts-16.20
- cb372/mu-scala-haskell-integration-tests:mu-haskell-0.4.0


serras (approved):
I updated the version of `mu-schema`, other than LGTM.

PR #1183, opened 2021-04-27
Update dependencies (avro4s update)

alejandrohdezma:
## What does this change do?

Updates all the updatable dependencies to their latest versions, except the ones that involve updating to cats-effect 3.0:

- Update Scala version (this involved a bunch of `nowarn` annotations).
- Update `cats-effect` to `2.5.0`.
- Update organization for `log4cats` (is now under the [Typelevel organization](https://github.com/typelevel/log4cats)).
- Update `avro4s` to `4.0.7` (this involved a few removed parameters and `Decoder` and `Encoder` classes now including the `SchemaFor` (instead of handling it implicitly).

## Checklist

### DIFF
--- a/benchmarks/shared/src/main/scala/higherkindness/mu/rpc/benchmarks/shared/ServerRuntime.scala
+++ b/benchmarks/shared/src/main/scala/higherkindness/mu/rpc/benchmarks/shared/ServerRuntime.scala
@@ -27,8 +27,8 @@ import higherkindness.mu.rpc.benchmarks.shared.server._
 import higherkindness.mu.rpc.server._
 
 import scala.concurrent.ExecutionContext
-import io.chrisdavenport.log4cats.Logger
-import io.chrisdavenport.log4cats.slf4j.Slf4jLogger
+import org.typelevel.log4cats.Logger
+import org.typelevel.log4cats.slf4j.Slf4jLogger
 import java.util.concurrent.TimeUnit
 import io.grpc.ManagedChannel
 import higherkindness.mu.rpc.channel.UsePlaintext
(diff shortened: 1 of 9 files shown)


fedefernandez (approved):
Looks good to me.

juanpedromoreno:
Yes, it seems that's the problem. That scala version causes `scoverage` issues like [this](https://github.com/scoverage/scalac-scoverage-plugin/issues/305). It's not exactly that, but I believe it's related.

I checked it in this PR: https://github.com/higherkindness/mu-scala/pull/1184, where the error is different (Avro test failing, apparently).

Thanks @alejandrohdezma !

franciscodr (approved):
Great work! Thanks, @alejandrohdezma

alejandrohdezma:
@juanpedromoreno downgrading Scala version again seemed to fixed the issue, but as you noticed the `RPCTests` failed :(

alejandrohdezma:
Can someone that knows their way around this tests help me troubleshoot it? There are so many logs that I'm a bit lost

alejandrohdezma:
@franciscodr managed to find the cause of the errors and the "solution". As said in https://github.com/higherkindness/mu-scala/pull/1183/commits/a232ee277d3551dde3c14d23782618b13ddf84e7, the only solution was to ignore the failing test. 

This test fails because of this update in `avro4s`: sksamuel/avro4s#594.

It is a corner case that will occur only if all the following applies:
- The user is generating IDLs from Scala code (it can never happen the other way around).
- A field in a request/response is using a `Coproduct`.
- Said `Coproduct` is using a generic type.
- Said field has a default value.

If all the previous checks, the generated schema will not contain the default value.

We discussed it and concluded that it is such a small case that we are safe to go down this road.

fedefernandez (approved):

juanpedromoreno (approved):

Issue #1328, opened 2021-10-07
Access Metadata in service 


ThatScalaGuy:
Hi,

right now, I am migrating a bunch of gRPC services to mu-scala. Those Services heavily use Metadata to store JWT tokens. 

Is there a way to access Metadata in the service?

fedefernandez:
Hi @ThatScalaGuy!

I would say that one of the best options is to create your interceptor. We have an example of using interceptors for the metrics. Take a look at how they are added to the server and the client here:

* https://higherkindness.io/mu-scala/guides/metrics-reporting

And then the source code of the implementation:

* https://github.com/higherkindness/mu-scala/blob/e38a8539f56019fd8b66010f998c84f876f4e652/modules/server/src/main/scala/higherkindness/mu/rpc/server/metrics/MetricsServerInterceptor.scala#L26

* https://github.com/higherkindness/mu-scala/blob/ccd8274d90c1979b4c7df035223cdaecd370bca3/modules/service/src/main/scala/higherkindness/mu/rpc/channel/metrics/MetricsChannelInterceptor.scala#L27

Let me know if you have any doubts.

ThatScalaGuy:
Hi, @fedefernandez 
Sorry for the late reply. 

Your suggestions were quite helpful for another issue I had.  But unfortunately, it does not solve the problem I have right now. 

I see no way to access the metadata directly in the service. 
I think it needs some adoption in the macro and code generation part. 

```
trait Greeter[F[_]] {
    def SayHello(req: HelloRequest): F[HelloResponse]
  }
```

Will become 

```
trait Greeter[F[_]] {
    def SayHello(req: HelloRequest, meta: Metadata): F[HelloResponse]
  }
```

Next week I will spend some afford to adopt your library. I am looking forward to finding an implementation that fits in the project's roadmap. 
Open for any discussion. Hopefully, my changes will be in the main branch soon

fedefernandez:
@ThatScalaGuy oh, I understand now. Yeah, at this moment is not possible. Your suggestion could make sense but the macro doesn't change the public interface. Another option is to have all the request objects extending from a type that contains the metadata. This is something we evaluated in the past.

ThatScalaGuy:
@fedefernandez hi, finally, I fixed all M1 issues, and I can start working on that topic. 

My approach is inspired by the akka grpc implementation I am currently using. They call it power API. In the end they have two traits - one with metadata and one without - 
[...]

PR #1408, opened 2022-01-12
Context service

fedefernandez:
This PR makes the `Span[F]` context abstract so that you can make your service dependant on any "context."

The idea comes from #1328

Pending tasks:
* [x] Add tests for a custom context
* [x] Update docs -> Will do in the following PR

Currently, we provide some methods in the companion object for getting a parameterized algebra over `Span[F]` with `Kleisli`

Given:

```scala
trait MyService[F]
```

The method `MyService.tracingClient[F[_]]` provides a:

```scala
Resource[F, MyService[Kleisli[F, Span[F], *]]]
```

And the method `MyService.bindTracingService[F[_]]` asks for an implicit instance implementation of the algebra parametized with `Kleisli[F, Span[F], *]` (`MyService[Kleisli[F, Span[F], *]]`).

In this PR, we're adding alternative methods to those "tracing" ones where you can specify any type as a context as soon you provide a couple of implicit implementations.

The new method `MyService.contextClient[F[_], C]` will require an implicit `ClientContext[F, C]`:

https://github.com/higherkindness/mu-scala/blob/9a0e04b85d7e3726596256b504daa1dfd82f3ee9/modules/service/src/main/scala/higherkindness/mu/rpc/internal/context/ClientContext.scala#L6-L15

While the new method `MyService.bindContextService[F[_]]` will require an implicit `ServerContext[F, C]`

https://github.com/higherkindness/mu-scala/blob/9a0e04b85d7e3726596256b504daa1dfd82f3ee9/modules/service/src/main/scala/higherkindness/mu/rpc/internal/context/ServerContext.scala#L6-L10

I've chosen `trait`s because the instances need to be parameterized over `F` and `C` but allow any `Req` and `Res` (this is passed in the `apply`). I'm open to suggestions for a more ergonomic or better way to provide these implementations.


AdrianRaFo (dismissed):

cb372:
LGTM. It's probably worth updating the docs (e.g. https://higherkindness.io/mu-scala/guides/distributed-tracing) and the [examples repo](https://github.com/higherkindness/mu-scala-examples/tree/master/tracing) to reflect these changes. And maybe add a small how-to guide to the docs, showing how to use this feature to access the metadata?

fedefernandez:
@cb372 yes! Initially, I was thinking about adding that in this PR, but this is getting bigger, and I prefer to keep this for code/implementation discussions and then open an issue to address that. Good point with the guide

modules/fs2/src/main/scala/higherkindness/mu/rpc/internal/client/fs2/calls.scala
```diff
@@ -86,72 +85,64 @@ object calls {
...
-            input.translate(Kleisli.applyK[F, Span[F]](span))
-          clientStreaming[F, Req, Res](
-            streamF,
-            descriptor,
-            channel,
-            options,
-            headers
-          )
-        }
+  )(implicit C: ClientContext[F, C]): Kleisli[F, C, Res] =
```
juanpedromoreno, line 93: _question_: would it be better if we rename this `C `implicit `val` differently from the `C[F]` implicit bound? (one or the other)

(same suggestion for the rest of the methods too)

juanpedromoreno (commented):
Besides my minor comment, shall we tag this PR as `breaking-change`?

fedefernandez:
@juanpedromoreno this should be backward compatible (besides the deprecated warning) so I think this should be fine without the `breaking-change` tag. I'll take a look at the val names.

juanpedromoreno (approved):

PR #1411, opened 2022-01-18
Update docs and adds a guide for accessing metadata

fedefernandez:
Depends on #1408 

This PR:
* Updates the tracing guide
* Adds a guide for handling/accessing metadata on services
* Adds a couple of utility methods for `ServerContext` and `ClientContext`


microsite/src/main/docs/guides/accessing-metadata.md
```diff
@@ -0,0 +1,237 @@
+---
+layout: docs
+title: Accessing metadata on services
+section: guides
+permalink: /guides/accessing-metadata
+---
+
+# Context on services
+
+Mu provides way to create contexts available in the client and/or in the server. Specifically, it provides the following features.
```
juanpedromoreno, line 10:
```suggestion
Mu provides a way to create contexts available in the client and server. Specifically, it offers the following features.
```

microsite/src/main/docs/guides/accessing-metadata.md
```diff
@@ -0,0 +1,237 @@
...
+
+## Client
+
+* For every RPC call, you need to create an initial context that will be passed to the client
+* The client will have the ability to operate and transform that context, that will be sent to the server in the headers.
```
juanpedromoreno, line 15:
```suggestion
* For every RPC call, you need to create an initial context that will be passed to the client
* The client will have the ability to operate and transform that context, which will be sent to the server in the headers.
```

microsite/src/main/docs/guides/accessing-metadata.md
```diff
@@ -0,0 +1,237 @@
...
+
+## Server
+
+* The server will have the ability to extract the context information from the request headers and use them.
+
+## How to use
+
+We'll assume the following service definition:
```
juanpedromoreno, line 23:
```suggestion
Let's assume the following service definition:
```

microsite/src/main/docs/guides/accessing-metadata.md
```diff
@@ -0,0 +1,237 @@
...
+
+@service(Protobuf, namespace = Some("com.foo"))
+trait MyService[F[_]] {
+
+  def SayHello(req: HelloRequest): F[HelloResponse]
+
+}
+```
+
+Let's look at how to enable the context on the client side first.
```
juanpedromoreno, line 39:
```suggestion
Let's look at enabling the context on the client-side first.
```

microsite/src/main/docs/guides/accessing-metadata.md
```diff
@@ -0,0 +1,237 @@
...
+
+To obtain a client with the context available, use `MyService.contextClient[F, C]` instead of
+`MyService.client`.
+
+This returns a `MyService[Kleisli[F, C, *]]`, i.e. a client which takes
+an arbitrary `C` as input and returns a response inside the `F` effect.
+
+This method requires an implicit instance in the scope. Concretely a `ClientContext[F, C]`
+
+```scala
```
juanpedromoreno, line 75: question: could we turn this into a mdoc block?

microsite/src/main/docs/guides/accessing-metadata.md
```diff
@@ -0,0 +1,237 @@
...
+      options: CallOptions,
+      current: C
+  ): Resource[F, ClientContextMetaData[C]]
+
+}
+```
+
+A `ClientContext` is an algebra that will take different information from the current call and the initial context (`current`)
+and need to generate a transformed context and a `io.grpc.Medatada`. The metadata is the information that will travel through
+the wire in the requests.
```
juanpedromoreno, line 95:
```suggestion
A `ClientContext` is an algebra that will take different information from the current call and the initial context (`current`)
and need to generate a transformed context and an `io.grpc.Medatada`. The metadata is the information that will travel through
the wire in the requests.
```

microsite/src/main/docs/guides/accessing-metadata.md
```diff
@@ -0,0 +1,237 @@
...
+
+There's a utility `def` in the companion object for generating a `ClientContext` instance from a function:
```
juanpedromoreno, line 97:
```suggestion
There's a `def` utility in the companion object for generating a `ClientContext` instance from a function:
```

microsite/src/main/docs/guides/accessing-metadata.md
```diff
@@ -0,0 +1,237 @@
...
+
+class MyAmazingService[F[_]: Applicative] extends MyService[F] {
+
+  def SayHello(req: HelloRequest): F[HelloResponse] =
+    HelloResponse(s"Hello, ${req.name}!").pure[F]
+
+}
+```
+
+Ordinarily, if you were not using context, you would create a gRPC service
```
juanpedromoreno, line 149:
```suggestion
In general, if you were not using context, you would need to create a gRPC service
```

microsite/src/main/resources/microsite/data/menu.yml
```diff
@@ -69,6 +69,10 @@ options:
     menu_type: guides
     url: guides/metrics-reporting
 
+  - title: Accessing metadata on services
+    menu_type: guides
+    url: guides/distributed-tracing
```
juanpedromoreno, line 74:
```suggestion
    url: guides/accessing-metadata
```

juanpedromoreno (commented):
Thanks @fedefernandez , this is great! I left some minor suggestions though :)

microsite/src/main/docs/guides/accessing-metadata.md
fedefernandez: This is already defined in the scope. Do you know if there's a way to embeed scala sources?

juanpedromoreno: Maybe you could use `mdoc:nest`? https://scalameta.org/mdoc/docs/modifiers.html#nest

fedefernandez: I tried, but that doesn't stop that code from providing new definitions to the scope. Honestly, I don't think this needs to be checked, it doesn't represent snippets that the users need to copy, just samples about how  the current code looks

fedefernandez:
@cb372 could you please take a look when you have a chance? Thanks!

microsite/src/main/docs/guides/accessing-metadata.md
```diff
@@ -0,0 +1,237 @@
...
+      channel: Channel,
+      options: CallOptions,
+      current: C
+  ): Resource[F, ClientContextMetaData[C]]
+
+}
+```
+
+A `ClientContext` is an algebra that will take different information from the current call and the initial context (`current`)
+and need to generate a transformed context and an `io.grpc.Medatada`. The metadata is the information that will travel through
```
cb372, line 94:
```suggestion
and generates a transformed context and an `io.grpc.Metadata`. The metadata is the information that will travel through
```

microsite/src/main/docs/guides/accessing-metadata.md
```diff
@@ -0,0 +1,237 @@
...
+}
+```
+
+To obtain a client with the context available, use `MyService.contextClient[F, C]` instead of
+`MyService.client`.
+
+This returns a `MyService[Kleisli[F, C, *]]`, i.e. a client which takes
+an arbitrary `C` as input and returns a response inside the `F` effect.
+
+This method requires an implicit instance in the scope. Concretely a `ClientContext[F, C]`
```
cb372, line 73:
```suggestion
This method requires an implicit instance in scope, specifically a `ClientContext[F, C]`:
```

microsite/src/main/docs/guides/accessing-metadata.md
```diff
@@ -0,0 +1,237 @@
...
+import io.grpc.{Metadata, MethodDescriptor}
+
+trait ServerContext[F[_], C] {
+
+  def apply[Req, Res](descriptor: MethodDescriptor[Req, Res], metadata: Metadata): Resource[F, C]
+
+}
+```
+
+Like in the case of the client, we have a `def` in the companion object that make us easier to build instances of `ServerContext`s:
```
cb372, line 192:
```suggestion
Like in the case of the client, we have a `def` in the companion object that makes it easier to build instances of `ServerContext`s:
```

cb372 (approved):
Looks good!

microsite/src/main/docs/guides/accessing-metadata.md
```diff
@@ -0,0 +1,237 @@
...
+import higherkindness.mu.rpc.server.{GrpcServer, AddService}
+
+object OrdinaryServer extends IOApp {
+
+  implicit val service: MyService[IO] = new MyAmazingService[IO]
+
+  def run(args: List[String]): IO[ExitCode] = (for {
+    serviceDef <- MyService.bindService[IO]
+    _          <- GrpcServer.defaultServer[IO](8080, List(AddService(serviceDef)))
+  } yield ()).useForever
```
cb372, line 164: TIL `useForever`

fedefernandez:
@juanpedromoreno ready for a second review

juanpedromoreno (approved):

PR #1448, opened 2022-03-29
Cross-build some modules for Scala 3

cb372:
* Drop Scala 2.12
* Cross-build these modules for 2.13 and 3: 
    * service
    * fs2
    * client-netty
    * client-okhttp
    * netty-ssl
    * server
    * prometheus
    * dropwizard

Inside the service module, the `@service` macro annotation and a few other things have moved to `src/main/scala-2/`

The remaining modules are not cross-built for a couple of reasons:

* they depend on something incompatible with Scala 3 (e.g. config module -> pureconfig -> shapeless)
* they depend on the `@service` annotation, e.g. the healthcheck module and the tests


juanpedromoreno (dismissed):
👍

fedefernandez:
Would it be possible to add Scala 3 to the build? Or is it already being cross-built?

fedefernandez:
I can't find the Scala 3 build in the CI, is that enabled for those modules?

fedefernandez (approved):

cb372:
Well spotted @fedefernandez. We can't run the tests for Scala 3 yet, but I've updated the `ci-test` command to at least compile the code against Scala 3.

cb372:
Argh, we can't drop Scala 2.12 because the sbt plugin has a dependency on it. I'll re-enable 2.12 just for the `mu-rpc-service` module.

cb372:
>Argh, we can't drop Scala 2.12 because the sbt plugin has a dependency on it.

Never mind, ignore this. The dependency from sbt-mu-srcgen to mu-rpc-service wasn't actually necessary so I'll remove it.

juanpedromoreno (approved):

PR #1453, opened 2022-04-05
Define our own Avro union types to replace shapeless Coproduct

cb372:
Define wrappers for Scala 3 union types, with custom instances of avro4s `SchemaFor`, `Encoder` and `Decoder` to encode them as Avro unions.

e.g. where in Scala 2 we would have used `Int :+: String :+: MyRecord :+: CNil` we can now use `AvroUnion3[Int, String, MyRecord]`.

I've added the unit test to the `service` module for now, which is inconsistent with our practice of collecting all tests in the `tests` module, because it's not possible to cross-build the `tests` module for Scala 3 yet. It has dependencies on a few Scala-2-only modules.

### DIFF
--- a/modules/service/src/main/scala-3/higherkindness/mu/rpc/avro/AvroUnions.scala
+++ b/modules/service/src/main/scala-3/higherkindness/mu/rpc/avro/AvroUnions.scala
@@ -7,6 +7,11 @@ import scala.reflect.Typeable
 
 final case class AvroUnion2[T1, T2](value: T1 | T2)
 final case class AvroUnion3[T1, T2, T3](value: T1 | T2 | T3)
+final case class AvroUnion4[T1, T2, T3, T4](value: T1 | T2 | T3 | T4)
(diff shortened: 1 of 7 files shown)


modules/service/src/main/scala-3/higherkindness/mu/rpc/avro/AvroUnions.scala
```diff
@@ -0,0 +1,567 @@
...
+          AvroUnion8[T1, T2, T3, T4, T5, T6, T7, T8](decoded)
+        } else if (canDecode[T8](schema8)(value)) {
+          val decoded: T8 = Decoder[T8].decode(schema8).apply(value)
+          AvroUnion8[T1, T2, T3, T4, T5, T6, T7, T8](decoded)
+        } else {
+          throw decodingException(value, schema1, schema2, schema3, schema4, schema5, schema6, schema7, schema8)
+        }
+      }
+    }
+
```
cb372, line 566: In case you're wondering, this file was not auto-generated, unless you count vim macros 😄

PR #1461, opened 2022-04-21
Rewrite the health-check module

### LINKED ISSUE #1054: Health check doesn't implement grpc health protocol
I'm trying to use health check https://github.com/higherkindness/mu-scala/issues/626 feature. But I can't figure out a way to check service status. 
I'm using https://github.com/grpc-ecosystem/grpc-health-probe/ cli program, but and when I run it, it says:
```
/grpc_health_probe-linux-amd64 --addr=0.0.0.0:50051 -v
parsed options:
> addr=0.0.0.0:50051 conn_timeout=1s rpc_timeout=1s
> tls=false
establishing connection
connection established (took 229.529469ms)
error: this server does not implement the grpc health protocol (grpc.health.v1.Health)
```
either on my own service:
```
      implicit0(hc: HealthCheckServiceFS2[F]) <- HealthServiceFS2.buildInstance[F]
      healthCheck                             <- HealthCheckServiceFS2.bindService[F]

      endpoints = List(
                   ...
                    AddService(healthCheck)
                  )
      server <- GrpcServer.default[F](r.config.rpcPort, endpoints)

      _ <- info"Gateway Starting"
      _ <- List(GatewayHttpEndpoint.launch(...),
                GrpcServer.server[F](server)
           ).parTraverse_(identity)
```

or on example (https://github.com/higherkindness/mu-scala-examples/tree/master/health-check) server:
```
sbt health-server-fs2/run
[info] Loading global plugins from /home/user/.sbt/1.0/plugins
[info] Loading settings for project mu-scala-examples-build from plugins.sbt ...
[info] Loading project definition from /home/user/IdeaProjects/mu-scala-examples/project
[info] Loading settings for project root from build.sbt,version.sbt ...
[info] Resolving key references (22777 settings) ...
[info] Set current project to mu-scala-examples (in build file:/home/user/IdeaProjects/mu-scala-examples/)
[info] running higherkindness.mu.rpc.healthcheck.sFs2.ServerApp 
```
Am I doing something wrong? Or it doesn't support "default" [healthcheck](https://github.com/grpc/grpc/blob/master/doc/health-checking.md)?

cb372:
We now use sbt-mu-srcgen to generate code from a .proto file, instead of defining the service in code with the macro annotation. That allows us to cross-build this module for Scala 3.

Because the `Health` service is generated from the official Protobuf definition taken from the [gRPC docs](https://github.com/grpc/grpc/blob/master/doc/health-checking.md), it's now compliant with the gRPC health check spec and thus works as expected with the standard gRPC tooling:

```
$ ./grpc_health_probe-darwin-amd64 -addr localhost:12345
status: SERVING

$ ./grpc_health_probe-darwin-amd64 -addr localhost:12345 -service A
service unhealthy (responded with "NOT_SERVING")

$ ./grpc_health_probe-darwin-amd64 -addr localhost:12345 -service B
status: SERVING

$ ./grpc_health_probe-darwin-amd64 -addr localhost:12345 -service C
error: health rpc failed: rpc error: code = NotFound desc =

```

Fixes #1054.

Also sneaking into this PR, I added cross-building for a couple more modules:

* `testing` (just a one-line change in build.sbt)
* `client-cache` (had to remove better-monadic-for)

This gets us closer to cross-building the tests module, which is the plan for the next PR.

## Regarding sbt-mu-srcgen

We now have a circular-ish dependency between the mu-scala and sbt-mu-srcgen repos:

* The scripted tests for sbt-mu-srcgen depend on the `mu-rpc-service` module in the mu-scala repo
* The `health-check` module in the mu-scala repo depends on sbt-mu-srcgen

I think at some point we should fold sbt-mu-srcgen into the mu-scala repo, because the two are so closely coupled.


modules/health-check/src/main/scala/higherkindness/mu/rpc/healthcheck/HealthService.scala
```diff
@@ -0,0 +1,97 @@
...
+     * https://github.com/grpc/grpc/blob/master/doc/health-checking.md
+     *
+     * "The server will immediately send back a message indicating the current
+     * serving status. It will then subsequently send a new message whenever
+     * the service's serving status changes."
+     */
+    val currentStatus =
+      Stream.eval(getStatus(req.service).map(_.getOrElse(ServingStatus.SERVICE_UNKNOWN)))
+    val futureStatuses = watchSignal.discrete
+      .filter(_.service == req.service)
```
fedefernandez, line 80: Can we use here triple equals from cats instead? Just in case we change the types in next changes

fedefernandez (approved):
This is great! Thanks

modules/health-check/src/main/scala/higherkindness/mu/rpc/healthcheck/HealthService.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  def build[F[_]: Concurrent]: F[HealthService[F]] = {
+    val checkRef: F[Ref[F, Map[String, ServingStatus]]] =
+      // empty string indicates the health of the server in general,
+      // rather than any particular gRPC service running on the server
+      Ref.of[F, Map[String, ServingStatus]](Map("" -> ServingStatus.SERVING))
+
+    val watchSignal: F[SignallingRef[F, ServiceStatus]] =
+      SignallingRef.of[F, ServiceStatus](
+        ServiceStatus("DummyService", HealthCheckResponse.ServingStatus.UNKNOWN)
```
fedefernandez, line 30: Where this comes from (the `DummyService` name)?

cb372: This is a bit of a hack, just because we have to initialize the `SignallingRef` with *something*.

I suppose the alternative would be to make it a `SignallingRef[F, Option[ServiceStatus]]`, initialise it with `None`, and wrap all service status updates in `Some(...)`. I'll try that and see how it looks.

cb372: Yep, that's a bit nicer.

fedefernandez (approved):

PR #1463, opened 2022-04-22
Cross-build the tests module

cb372:
* Remove dependency on config module, as that's not cross-built
	* Move the relevant tests to the config module
	* Also improve one of the tests by adding a missing config file
* Move all uses of the `@service` macro to `src/test/scala-2/`
* Minor syntax changes to satisfy the Scala 3 compiler

The next step will be to gradually rewrite the uses of `@service` in the tests to use sourcegen instead. But that will require some thought, as the tests use many combinations of serialization type (Avro/AvroWithSchema/Protobuf), compression (Gzip/Identity) and idiomatic endpoints (true/false). It's not immediately clear how we can support that, as sbt-mu-srcgen only lets you configure those on a per-project basis.


modules/tests/src/test/scala/higherkindness/mu/rpc/dropwizard/DropwizardMetricsTests.scala
```diff
@@ -62,8 +62,6 @@ class DropwizardMetricsTests extends ScalaCheckSuite {
           (registry.getTimers().asScala.toMap, registry.timer(gaugeName).getCount)
         case Counter =>
           (registry.getCounters().asScala.toMap, registry.counter(gaugeName).getCount)
-        case _ =>
```
cb372, line 65: The Scala 3 compiler helpfully pointed out that this case is unreachable!

modules/config/src/test/resources/application.conf
```diff
@@ -0,0 +1 @@
+rpc.server.port = 51111
```
fedefernandez, line 1: where is this used @cb372 ?

fedefernandez (approved):

modules/config/src/test/resources/application.conf
cb372: In this test: https://github.com/higherkindness/mu-scala/pull/1463/files#diff-0ad51a2d7069328bafc89a9c1cfc566aee3f1e30a8d3a55be6912b17f246af39R32-R40

PR #1464, opened 2022-04-26
Port a bunch of tests to use source generation instead of the macro annotation

cb372:
This allows the tests to run against both Scala 2 and Scala 3.

Thanks to these tests, I discovered that the Avro enum encoding I chose for Scala 3 (sealed trait + case objects) is not actually supported yet by Avro4s v5.x 🤦 . Schema generation works, encoding works, but decoding throws a `NotImplementedError`. I've added a test to record this, and submitted a fix upstream.


juanpedromoreno (approved):
👌

PR #1468, opened 2022-04-27
Switch to sbt-crossproject for cross-building

cb372:
We have most modules targeting Scala 2.13 + Scala 3, a few modules (`config`, `benchmark-vnext`, `haskell-integration-tests`) that only target 2.13, and now I want to upgrade the microsite to target Scala 3.

Unfortunately that combination seems to be impossible using sbt's built-in `scalaVersion`/`crossScalaVersions` mechanism. Whatever I do, I get errors about both `_2.13` and `_3` versions of dependencies being on the classpath.

Switched to sbt-crossproject, which makes it really easy to handle situations like this. For each module, the plugin generates separate synthetic projects for each Scala version, and sets up the inter-project dependencies appropriately.

It also allows more parallelisation, as the 2.13 and 3 versions of a module will build in parallel.

This PR also includes a little housekeeping:
* Removed sbt-scoverage and sbt-codecov, as we weren't using them for anything
* Removed some test certificate files that weren't being used
* Refactor the tests to choose a fresh server port for every test case, as I noticed some occasional test flakiness caused by re-use of the same port across tests
* Fix `.scalafmt.conf` so we can format Scala 3 sources correctly


juanpedromoreno (approved):
Great, thanks @cb372 !

